PR #1, opened 2022-01-04
Implement `Display` for lambolt definitions.

nothingnesses:


MaiaVictor:
Wow, thank you. It is ok to keep it recursive for now.

PR #3, opened 2022-01-05
Add GitHub actions

nothingnesses:
Formats rust code and runs clippy linting on pushes and pull requests.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,0 +1,14 @@
+on: [push, pull_request]
+name: Clippy
+jobs:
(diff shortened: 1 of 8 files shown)


rigille:
Thanks @nothingnesses! I was wondering how to do that just now

nothingnesses:
Ah I just noticed [there's bug reports](https://github.com/actions-rs/clippy/issues) saying that the clippy GH action passes even when the clippy lints fail. Maybe we should [replace it with this](https://github.com/actions-rs/clippy-check) for now until that bug is fixed?

PR #7, opened 2022-01-10
Use ropes for stringifiers.

nothingnesses:
Tbh, I'm not sure this would be faster than the previous implementation, since there's now conversions occurring between strings and ropes. I think it could probably be made faster if the amount of conversions are minimised and if all the interpolations are only done with ropes, and conversion to strings only happens once after these are done. Alternatively, maybe we shouldn't rely on the `Display` trait and `write` macro for interpolation?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,6 +146,7 @@ dependencies = [
  "cranelift-native",
  "lazy_static",
  "regex",
+ "ropey",
 ]
 
 [[package]]
@@ -222,6 +223,15 @@ dependencies = [
  "libc",
  "mach",
  "winapi",
+]
+
+[[package]]
+name = "ropey"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e6b9aa65bcd9f308d37c7158b4a1afaaa32b8450213e20c9b98e7d5b3cc2fec3"
+dependencies = [
+ "smallvec",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ cranelift = "0.80.0"
 cranelift-jit = "0.80.0"
 cranelift-module = "0.80.0"
 cranelift-native = "0.80.0"
+ropey = "1.3"
 
 [profile.dev]
 # opt-level = 0
--- a/src/lambolt.rs
+++ b/src/lambolt.rs
@@ -1,5 +1,7 @@
 use crate::parser;
+use ropey::Rope;
 use std::fmt;
+use std::fmt::Write;
 
 // Types
 // =====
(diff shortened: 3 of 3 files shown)


rigille:
Cool idea, I was also thinking about using RRBs

quleuber:
I'm not sure this library helps a lot with stringfication, because we are not actually editing the strings just building them(?)

Maybe we could do this efficiently just chaining iterators?

@nothingnesses what do you think?

nothingnesses:
@quleuber

> I'm not sure this library helps a lot with stringfication, because we are not actually editing the strings just building them(?)

True. Tbh, I just tried to use it because @MaiaVictor [left a note](https://github.com/Kindelia/LambdaVM-Rust/blob/2a7e01b191a31d7c569528f431953800a3fcaa14/src/lambolt.rs#L119) saying ropes should be used instead (of `String`s?).

> Maybe we could do this efficiently just chaining iterators?

I'm not sure I understand what you're getting at exactly. How should iterators be used in the context of building strings from the structs? Are you saying we should implement the `Iterator` trait for `Oper`, `Term`, `Rule` and `File`, and then iterate through these with a `for_each` that pattern matches against the variants, and returns the corresponding string slices, then `collect` these all into the finished `String`? If not, please could you explain what you meant?

I'm also not sure if using iterators would be more efficient than previous implementations, but I should probably mention that [chaining together iterators is apparently slower than using a single iterator](https://nnethercote.github.io/perf-book/iterators.html#chaining), so maybe we should try to minimise doing it if possible.

MaiaVictor:
@quleuber can we merge this?

@nothingnesses are you working in some file currently? I was thinking in finishing readback today, if you're not editing it.

quleuber:
@nothingnesses 
I was thinking something like:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6beb6ddd58cb80670691fde03333fad8

Now that I wrote it if feels it has a terrible overhead, indeed.  :P

nothingnesses:
@MaiaVictor I'm not working on readback currently, please go ahead with editing it.

nothingnesses:
Closing this as I'm pretty sure this just ends up doing doubling the memory allocations (one for the `SmallVec` used by the `RopeBuilder` and one for the output string) compared to the current solution and so is likely slower.

VictorTaelin:
Wait, really? Does that apply to very big strings? That doesn't make sense from an asymptotic point of view. How could?

nothingnesses:
@VictorTaelin Tbh I'm not completely sure. I'd have to run some benchmarks to be more confident about what I said. But essentially, when I said what I said earlier, I was making an assumption that the memory used by the rope won't

nothingnesses:
@VictorTaelin So [I wrote a benchmark](https://nest.pijul.com/nicoty/rope_bench) to test if it's faster to build up `String`s directly, or to use `RopeBuilder` to do it (run it with `./bench` in the terminal). Granted, the test cases are pretty small and according to ropey's readme, it works better with larger strings, but in any case, these are the results I got:
```
$ ./bench
   Compiling rope_bench v0.1.0 (/home/a/tmp/rope_bench)
    Finished release [optimized] target(s) in 10.81s
     Running `target/release/examples/criterion --bench`
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

Gnuplot not found, using plotters backend
Formatters/StringContainerA/
                        time:   [2.7793 us 2.7837 us 2.7889 us]
Formatters/StringContainerB/
                        time:   [10.112 us 10.136 us 10.159 us]
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe
```
`StringContainerA` builds up the `String` directly, whereas `StringContainerB` uses `RopeBuilder` under the hood. According to the result, at least with smaller strings, it looks like using ropey would be slower with the formatter implementation in this PR.

I don't know a good way to generate larger strings though so we can test with those too, maybe it would be much faster with much larger strings.

PR #12, opened 2022-01-18
remove let and name from unused variables after sanitize

racs4:
Also fix some lint errors and add unit tests for `compilable.rs`


VictorTaelin:
Thank you!

PR #14, opened 2022-01-23
Add LICENSE

nothingnesses:


VictorTaelin:
It has some conflicts though

nothingnesses:
Ah sorry. The only change I made was adding a license file, but the rustfmt bot formats the other files automatically. I rebased from current master now though.

VictorTaelin:
Okay, thank you. Are you or someone else planning to use HOVM for something? :)

stken2050:
Hi, in terms of the naming, I feel "Virtual Machine" does not suit the shining concept of non-garbage-collected (like Rust).
VM runtime easily reminds us of JAVA with GC, and nowadays people want to avoid it more than before.

Probably as the author you have some idea behind the naming, but at least mystery for me, and probably the same goes for anyone who is interested in this runtime.

leonardohn:
I don't think virtual machines have anything to do with garbage collection, although it is a common practice to virtualize memory access. The thing is, even though HOVM is capable of generating binaries, it always carries runtime code to reduce the lambda terms into their normal form, which can be seen as an embedded virtual machine that executes lambda-encoded programs.

stken2050:
Yeah, actually, I've been confused by the concept of the project, but now I understand this will be the foundation of Kindelia that is currently built on TypeScript/Deno/Kind, also I've noticed the word VM is strongly inspired by Ethereum Virtual Machine (EVM), so my concern on the impression of VM is wiped out especially by the power word of EVM.
I'm sorry for disturbing the issue.

VictorTaelin:
Thanks for bringing these concerns, I'm really undecided about the name. The VM comes for the fact the runtime is based on constant-time operations (like OPCODES) that modify the memory. So, for example, BETA_REDUCE is a primitive opcode in HOVM, just like ADD is an opcode on the EVM, WASM, LLVM IR, etc. 

But as you said, I don't think the name captures what it does that well. Calling it a "runtime" (LamRT? HORT?) may be cool, but not many things are called runtimes. Calling it a language sounds technically right, but it intends to be more of a "low-level" (that is actually high-level) compile target than a fully featured user-facing language.

Naming this project is hard because there isn't anything quite like it, but I'm still accepting suggestions.

nothingnesses:
What about calling it a "compilation target" instead of "virtual machine"? So maybe something like "Higher Order Compilation Target" or "Functional Compilation Target"?

stken2050:
Please add the term "core" in the list, personally, I like the wording like Formality-Core.

It explains well :)

leonardohn:
But a target is in fact a machine. Take for instance the LLVM compilation targets like X86, ARM, SPIR-V or even WASM; They all resemble computer architectures that are either real hardware (as in X86 and ARM) or a virtual hardware (like SPIR-V or WASM). I agree that historically virtual machines are usually related to traditional hardware (sharing characteristics like using von Neumann-like architectures), but in this particular case, we are talking about an abstraction that doesn't have memory at all, just like original lambda calculus when compared to Turing machines. So, it doesn't even make sense talking about garbage collection inside this machine, unless when considering the implementation of this abstraction over bare metal hardware, which in turn also don't require garbage collection.

nothingnesses:
@leonardohn True. Personally I already quite like "HOVM", I was just spitballing some other ideas since @stken2050 mentioned that "VM" reminds them of GC.

stken2050:
I also think HOVM is totally fine. Probably I should've not mentioned JVM stuff, I was just personally very confused. I really like @leonardohn comment above.

VictorTaelin:
... conflicts

PR #18, opened 2022-01-25
Add profiler.

nothingnesses:
Todo:
 - [ ] Add output generation dependencies to dev shell.
 - [ ] Add a CLI to parameterise input programs.

Not sure how useful this will be, in any case, here's some example outputs:

flamegraph

profile001

They're supposed to show which parts of the code the programs spends the most time on. E.g., with the above outputs, they show that least 25% of running time was spent on calls to `builder::alloc_body` so we might want to try to further optimise that part of the code.

It's a bit weird that it only seems to show one path through the code though (I don't see any of the other functions that should be making up the other 75% of the running time). Not sure if that's because my code is whack (this is most likely), or due to some limitation of pprof-rs, or because of the input hovm code I used.

Edit: Ah it seems like uploading to GitHub broke the interactivity of the SVGs. Well anyway, they're also supposed to be interactive...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,7 @@ cranelift-native = "0.80.0"
 
 [profile.release]
 opt-level = 3
+
+[profile.bench]
+opt-level = 3
+debug = 2
--- a/profiler/Cargo.toml
+++ b/profiler/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "profiler"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+hovm = { path = "../" }
+pprof = { version = "0.4", features = ["flamegraph", "protobuf"] }
+
+[profile.bench]
+opt-level = 3
+debug = 2
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -1,0 +1,18 @@
+Profiling
+---------
+
+#### 1. Install Dependencies
+
+These are optional dependencies to generate the flowchart output (the flamegraph output doesn't require them):
+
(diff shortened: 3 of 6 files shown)


VictorTaelin:
Can it avoid changing the formatting of unrelated files?

nothingnesses:
I tried looking for such an option, but I couldn't find one. Should I just make a PR removing the bot for now? I'll look for alternatives in the mean time.

VictorTaelin:
I'm not sure if the profiler should be added on this repo though, shouldn't that be a separate project? Is that usual on Rust projects?

nothingnesses:
Not sure about that either, but I guess I haven't seen a lot of projects that also include profiling harnesses, so maybe not. I guess it doesn't strictly need to be included in this repo, could just be on a repo of its own.

nothingnesses:
I think the other functions weren't being displayed because they were being inlined. I'll try and figure out a convenient way to prevent that for all functions when profiling so that the profiling outputs are more useful.

nothingnesses:
Here are the new outputs after my recent pushes:
flamegraph
profile001
They're have a bit more detail now, so hopefully they'll be more useful in determining which functions we should be trying to optimise. Uploading to GH still messes up the SVGs though, so you'll have to run them yourself.

quleuber:
> I tried looking for such an option, but I couldn't find one. Should I just make a PR removing the bot for now? I'll look for alternatives in the mean time.

@nothingnesses

You can disable the bot/automation without removing the config:
https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

in this page, I guess:
https://github.com/nothingnesses/LambdaVM-Rust/actions

VictorTaelin:
I think making the functions public and leaving it for another repo makes more sense, but I'm not sure. @quleuber thoughts?

PR #21, opened 2022-01-26
fix C compilation warnings on Linux

quleuber:
This fixes these warnings while compiling on Linux:

```
main.hovm.out.c:110:8: warning: variable length array folded to constant array as an extension [-Wgnu-folding-constant]
  Stk  free[MAX_ARITY];
       ^
main.hovm.out.c:129:8: warning: variable length array folded to constant array as an extension [-Wgnu-folding-constant]
Worker workers[MAX_WORKERS];
       ^
main.hovm.out.c:1160:5: warning: variable length array folded to constant array as an extension [-Wgnu-folding-constant]
u64 normal_seen_data[normal_seen_mcap];
    ^
main.hovm.out.c:1334:23: warning: incompatible pointer to integer conversion assigning to 'Thd' (aka 'unsigned long') from 'void *' [-Wint-conversion]
    workers[t].thread = NULL;
 ```

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -479,12 +479,12 @@ const u64 U64_PER_MB = 0x20000;
 const u64 U64_PER_GB = 0x8000000;
 
 #ifdef PARALLEL
-const u64 MAX_WORKERS = 8;
+#define MAX_WORKERS (8)
 #else
-const u64 MAX_WORKERS = 1;
+#define MAX_WORKERS (1)
 #endif
 const u64 MAX_DYNFUNS = 65536;
-const u64 MAX_ARITY = 16;
+#define MAX_ARITY (16)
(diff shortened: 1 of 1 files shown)


VictorTaelin:
Isnt the =NULL needed there?

quleuber:
The `= NULL` assignment isn't needed.

The `.thread` field is of type `pthread_t`. From [these docs](https://www.ibm.com/docs/en/i/7.1?topic=ssw_ibm_i_71/apis/concep17.htm):

> the pthread_t abstract type

> Do not allow your program to rely on the internal structure or size of the pthread_t in a non-portable fashion, such as comparisons of thread IDs.

So, this value should not be manipulated directly. The field is initialized right after by `pthread_create` on [line 1367](https://github.com/quleuber/HOVM/blob/0247e021e69a8a1bb17caefd269dd43f2e75f95b/src/compiler.rs#L1367).

PR #22, opened 2022-01-27
Improve README and some code.

nothingnesses:


nothingnesses:
No worries, just glad to help! If I were as much of an expert at type theory as you are, I think I wouldn't mind trading a few of my English skills for it. Type theory is at least a more universal and less arbitrary language than English.

PR #28, opened 2022-01-27
Improve README. Add Matrix bridge link.

nothingnesses:


VictorTaelin:
SHouldn't we encourage people to join the server?

nothingnesses:
Well the Matrix bridge connects to the Discord server and vice versa, so you can access the messages from either platform if you join either one. The Matrix bridge is essentially just a convenient way for people to access the Discord server since they don't have to have a lot of messaging apps on their device to be able to communicate with people on different platforms (i.e., I just use my Matrix client to communicate with people on Telegram and Discord, so I don't have to visit their separate apps, I just access everything from Element).

VictorTaelin:
Yes but I guess that's kinda against the spirit of Discord communities, no?

nothingnesses:
I'm not sure I understand what you mean, would you care to elaborate?

VictorTaelin:
Discord communities are fun because everyone is able to join, get roles, participate on voice chats, etc. You don't have that if you join on Matrix, right?

nothingnesses:
Ah crap, the Discord link broke. It seems I doubled the parentheses around the link, sorry.

PR #31, opened 2022-01-27
Bunch of changes.

nothingnesses:
- [x] Check and test Rust code.
- [x] Only check Rust formatting instead of applying it.
- [x] CI for Nix build.
- [x] Eliminate CI failures.
- [x] Remove unused dependencies in Cargo.nix.
- [x] Fix flake.
- [x] Improve TL;DR.
- [ ] Check and test C outputs.


VictorTaelin:
Should we fix these unsuccessful checks before merging?

nothingnesses:
They're just the clippy lints and formatting checks. I think before we fix the formatting failures, it might be a good idea for you to peruse the [rustfmt documentation](https://rust-lang.github.io/rustfmt/), review the actual rustfmt.toml configuration we currently have and make changes to it to your liking so that we can have a consistent "house style" for the whole project that every contributor can agree to stick to.

VictorTaelin:
Okay, I'll do that when I get back to work

VictorTaelin:
Okay, that is huge. I don't really have strong opinions, I agree with whatever the Rust community finds good, other than tabs being 2 spaces.

nothingnesses:
That's fine. The good thing about having a formatter is we can just edit the configuration later on and have it automatically re-format the code later on, so we don't have to stick to one style forever. So in that case, if you're happy with the current config, I can try fixing the causes of the CI failures and update this so that this can finally be merged?

nothingnesses:
@VictorTaelin I've eliminated the CI failures. I re-denied some previously globally allowed lints, making some of them locally allowed instead (i.e., mostly adding the `#[allow(dead_code)]` attribute above a couple of functions) or actually replacing the code that clippy was complaining about (i.e., mostly adding underscores before unused variables). Please feel free to review whenever.

nothingnesses:
> Okay, that is huge. I don't really have strong opinions, I agree with whatever the Rust community finds good, other than tabs being 2 spaces.

Wait, I just re-read your comment and I'm not sure I interpreted it correctly. Just to be clear, are you saying that the tabs *should* be 2 spaces, or *shouldn't* be 2 spaces? If it's the latter, please could you let me know how many spaces it should be, instead of 2?

VictorTaelin:
Can it not touch the hadouken?

nothingnesses:
> Can it not touch the hadouken?

I think we could use `#[rustfmt::skip]` for that. Let me check.

PR #32, opened 2022-01-28
compiler: switch to token replacement instead of format strings

lqd:
This allows the use of regular `{` and `}` tokens in the C file instead of the escaping required for every scope and control flow structure. It should be a QOL improvement while developing the C file, compared to the cost of a few string allocations while compiling to C ?

Also a couple of drive-by changes: 
- removes some, I believe unused, cranelift crates: hvm will be faster to build
- some automated rustfmt, and println simplifications

(I haven't fully tested this, only on one example, which seemed to work. Hopefully CI will soon be able to test this PR)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,18 +3,6 @@
 version = 3
 
 [[package]]
-name = "anyhow"
-version = "1.0.53"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "94a45b455c14666b85fc40a019e8ab9eb75e3a124e05494f5397122bc9eb06e0"
-
-[[package]]
-name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
-
-[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


nothingnesses:
> (I haven't fully tested this, only on one example, which seemed to work. Hopefully CI will soon be able to test this PR)

Ah, unfortunately the CI doesn't yet run automated tests, sorry. It should be added when #31 gets merged though. We also don't have extensive unit tests yet, so I'm not sure if it'll be able to catch a lot of errors either when it does. I think I might try integrating some property testing with the proptest crate to improve this.

lqd:
I was indeed thinking of your PR being able to test this PR soon. Even basic smoke tests will be able to catch an intended change here.

lqd:
Conflicts on `runtime.c` are super likely since it's an heavily modified file. If that happens, I'll rebase when/if people find this technique interesting.

quleuber:
Since we are still loading the template on compile time, I think this is better. Let's see what Maia thinks.

I just wish there was a function that would do all the replacements concurrently.

lqd:
Or you could search the template and split it per token "hole", write each static section directly to the output file without doing interpolation and the String allocation & concatenation. In any case, this is easy to improve when/if you see that generating the C file becomes the bottleneck of the compiler, which seems unlikely to be the case today ?

VictorTaelin:
This looks good. I won't be updating the code a lot tomorrow, so that's a good window to merge

lqd:
Rebased.

I've left the tokens independent, so that you can change them individually. It will be easy to make the template a valid C file to ease development by turning into valid commented expressions to replace.

nothingnesses:
@lqd Ideally, you should run `crate2nix generate` if you're able to install `crate2nix` (or if you have Nix, you can access it from the dev shell) since your PR changes Cargo.lock. But don't worry though, I've included the required changes in my CI PR instead so you don't have to do that.

lqd:
@nothingnesses thank you for doing that, I don't use nix.

quleuber:
I guess to make 
```c
#define MAX_WORKERS (/* GENERATED_NUM_THREADS_CONTENT */)
```
we could do something like
```c
/* GENERATED_NUM_THREADS_CONTENT */
#ifndef MAX_WORKERS
#define MAX_WORKERS (1)
#endif
```
`/* GENERATED_NUM_THREADS_CONTENT */` <- `#define MAX_WORKERS ({num_threads})`

etc
?

lqd:
Sure, or variations of `int value = /* START */ 0 /* END */;` expressions where it is important, and replace both tokens and the content they box.

PR #34, opened 2022-01-30
Improve and add tests for parser.

nothingnesses:
Ok so the major changes in this PR so far are:
* I removed some functions that seemed unused/unneeded:
  * `equal_at`, whose usages I replaced with `rest`.
  * `flatten`, which is just the `concat` method for slices.
  * `lines`, which was kinda bad anyway, as it unnecessarily allocated memory.
  * `head_default`, no longer needed due to changes to `skip_spaces`.
* `skip` no longer recurses, which should prevent that function from causing any stack overflows. It also mutates its argument now (not sure if that's bad, doesn't seem like it affects anything though).
* Similarly, the function signatures of `skip_comment` and `skip_spaces` now explicitly state that the argument is mutated (it seems like they were previously implicitly mutating their arguments anyway?).
* The `rest` method now doesn't directly index into a slice (which is dangerous and can cause panics), it instead calls `get` and returns an Option.
* `highlight` should allocate less memory now (only the allocation of the intermediate `String` should be occurring now, instead of also allocating a `Vec`).
* Added tests for a few of the functions (`tail`, the skipper functions, and `highlight`).

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -26,6 +26,7 @@
 // Check the Testree example at the bottom of this file.
 
 #![allow(dead_code)]
+use itertools::Itertools;
 
 // Types
 // =====
@@ -37,8 +38,8 @@ pub struct State<'a> {
 }
 
 impl<'a> State<'a> {
-  fn rest(&self) -> &'a str {
-    &self.code[self.index..]
+  fn rest(&self) -> Option<&'a str> {
+    self.code.get(self.index..)
   }
 }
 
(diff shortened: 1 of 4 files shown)


VictorTaelin:
Looks great! But have you tested the new highlight for different inputs? One line, zero lines, 10+ lines, errors covering 1 and multiple lines?

nothingnesses:
Not yet. I'm in the process of writing some property-based tests for functions in parser.rs, which I'm writing in conjunction with this (but that will be another PR), which is why this is still marked as WIP, since I might discover other things to improve in that file while writing the tests and getting a better understanding of its functions.

VictorTaelin:
But are you sure it is the old algorithm? Any chance something changed unnoticed? Just to be sure before merging.

nothingnesses:
It's entirely possible I broke something. I'll definitely do the tests you asked and review any changes before I mark this ready for review.

src/parser.rs
```diff
@@ -379,63 +373,72 @@ pub fn highlight(from_index: usize, to_index: usize, code: &str) -> String {
...
+  for (i, c) in code
+    .chars()
+    .enumerate()
+    .filter(|(_, c)| c == &'\n')
+    .take_while(|(i, _)| i < &to_index) {
       if i < from_index {
         from_line += 1;
       }
       if i < to_index {
         to_line += 1;
```
nothingnesses, line 385: TODO: Remove this redundant check.

VictorTaelin:
okay, just let me know when you're confident it is equivalent, and I'll merge!

nothingnesses:
I want to take a break from working on the parser (I'm not sure what other easy improvements I can make to it other than adding more tests) and to instead try to dabble with other parts of the repo, so I've marked this ready for review so that the work I've done so far at least won't be wasted.

VictorTaelin:
I've been short of time to review, test and merge PRs, sorry. Please just notify me via DM on Discord when you need me so merge something. Can't merge this one due to conflicts sadly.

PR #37, opened 2022-01-31
Hadouken.

nothingnesses:
Addresses this grave concern:

> Can it not touch the hadouken?

_Originally posted by @VictorTaelin in https://github.com/Kindelia/HVM/issues/31#issuecomment-1025938142_


VictorTaelin:
Hadouken.

PR #50, opened 2022-02-01
Add Infinite lists sample codes

stken2050:
https://www.reddit.com/r/rust/comments/sh947t/hvm_a_nextgen_massively_parallel_betaoptimal/hv2dxm8/?utm_source=reddit&utm_medium=web2x&context=3


VictorTaelin:
This feels a little bit out of place there. Perhaps we should have an examples directory (or file)?

stken2050:
That would be cool. Thank you.

PR #63, opened 2022-02-03
Improve CI and fix stuff.

nothingnesses:
Addresses:
> Also, it would be great if we could test in multiple platforms/compilers on CI. @nothingnesses any ideas?

_Originally posted by @quleuber in https://github.com/Kindelia/HVM/issues/62#issuecomment-1029216699_


nothingnesses:
Please review @quleuber or @VictorTaelin

PR #67, opened 2022-02-06
Correct tB to yB

kconner:
This change fixes what looks like a mismatch in terms between a reduction step and its written explanation.


VictorTaelin:
Thanks!

Issue #53, opened 2022-02-02
`runtime.c` template as valid C file


quleuber:
Right now the `src/runtime.c` template is not a valid C file before the substitutions are made. This hardens developing on that file unnecessarily.

Some solutions to this are discussed on #32, e.g.

> Sure, or variations of int value = /* START */ 0 /* END */; expressions where it is important, and replace both tokens and the content they box.

VictorTaelin:
That's clever.

cauefcr:
https://github.com/Kindelia/HVM/pull/68 🙏

quleuber:
Fixed by 
e08f17e91cba65f9a94d0bbbc3365b0c7dd2c264
a9bb0ff0797301d85788034a2cc74ee8c153ffa0

PR #68, opened 2022-02-07
Makes a parallel sum the standard program in case the runtime is compiled by itself

cauefcr:
Saw the issue and decided to fix it.

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -513,7 +513,7 @@ pub fn c_runtime_template(
   // Instantiate the template with the given sections' content
 
   C_RUNTIME_TEMPLATE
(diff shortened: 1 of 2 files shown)


VictorTaelin:
Not sure I get it, that means every compilation will jet injected the FIB function, even if unrelated? 

Has `--single-thread` been removed?

quleuber:
> expressions to be replaced like in the comment I quoted

I'll implement this.

quleuber:
I closed the original issue (#53) by other means. Now we just have to decide if we keep a default program on `runtime.c` or not.

VictorTaelin:
Hi, any updates on the state of the review of this PR? Can it be merged?

PR #70, opened 2022-02-08
Replace "Cargo.nix" with `generatedCargoNix`.

nothingnesses:
With Nix 2.6.0 being released, it's no longer necessary to manually generate "Cargo.nix" as the automatically generated `generatedCargoNix` derivation can now be used in "flake.nix" without resulting in errors.


nothingnesses:
It's weird that the CI didn't run. Maybe I broke something with one of my previous commits? This PR [failed one of the workflows though](https://github.com/nothingnesses/LambdaVM-Rust/runs/5117288388?check_suite_focus=true), so please don't merge yet.

nothingnesses:
Huh, I had to manually mark it for review before the CI fired. That's no good. I'll try to fix that, as well as figure out the cause of the failing check.

balsoft:
The CI failure is, unfortunately, quite expected. Since `generatedCargoNix` is IFD, it requires to be able to build for each platform the CI evaluates for, which is not possible (without remote builders at least). One solution is to do `nix build -f . checks.x86_64-linux` instead of `nix flake check`.

(Yes, Nix devs aware of this issue and there will hopefully be a solution one day)

nothingnesses:
@balsoft Building the checks attribute for a specific system architecture like you said did the trick, thanks! Presumably, [this is the issue](https://github.com/NixOS/nix/issues/4265) I should track if I want to know if there's been any changes regarding this, right?

PR #73, opened 2022-02-09
Do misc cleanup

azdavis:


src/main.rs
```diff
@@ -16,31 +16,32 @@ fn run_cli() -> std::io::Result<()> {
...
-  if args.len() <= 1 {
-    show_help();
-    return Ok(());
-  }
-
-  let cmd = &args[1];
+  let cmd = match &args[..] {
+    [] | [_] => {
+      show_help();
+      return Ok(());
```
nothingnesses, line 29: I think you can get rid of the `return` here?

nothingnesses:
Strange that clippy didn't catch the redundant `return`s and didn't recommend using the `matches!` macro. I wonder why that was?

PR #75, opened 2022-02-09
Fix <= parsing

LeoRiether:
As of v0.1.7, [`parse_op2` parses `<` before `<=`](https://github.com/Kindelia/HVM/blob/master/src/language.rs#L311-L312), which implies the term `(<= 1 2)` doesn't parse correctly -- this fails with `thread 'main' panicked at 'No parse.'` because it thinks it's a `<` operator followed by something it didn't expect).

I just changed the order of `Oper::Ltn` and `Oper::Lte` to fix this :)


VictorTaelin:
Thanks!

PR #79, opened 2022-02-11
add ToC and "Building" section on README

quleuber:


nothingnesses:
The TOC is a good idea, but having the  "Building" way separate from the installation (or I guess "Usage") section seems a but weird. Should maybe be under "Usage"? The Nix sub-section also seems redundant as it's already been linked in "Usage", but I guess it makes some sense. Could probably also do without "from sources" in the instructions, since they're already under the "Building" heading, so I think it's fairly safe to assume that they imply "from sources", even if they get removed. Other than that, looks good! 👍

PeterNjeim:
I agree with @nothingnesses, except for the Building section being at the bottom, since most GitHub pages I've seen have the Building or Contributing section at the bottom, mainly because the majority of users don't intend on building, and instead want to use and learn about the program. In the end it doesn't really matter where it's placed though, due to the table of contents.

nothingnesses:
True, I guess having "Building" close to "Contributing" makes some sense. TOC does help, but I'd say having a coherent structure to documentation is important too. It'd be stupid to have the introduction to a project placed at the end of the documentation, for instance.

Issue #102, opened 2022-04-07
Black box testing


quleuber:
Python script to run blackbox tests on HVM programs.
e.g.
https://github.com/quleuber/HVM/tree/tests/test

PR #108, opened 2022-04-09
Add black-box test script + better workflows

AdrianoII:
Creates a sub-directory and a simple Python script to run black-box tests #102.

I try to only use libs from the Python standard library to avoid adding the complexity of managing Python dependencies.

And I also try to build the script in a way that it can be called from anywhere, so calling from the root of HVM should be possible.

Each test is a directory inside `tests/BlackBox` and contains an HVM source code and a JSON that contains the tests cases (inputs and expected outputs).

So the script will run each test case from each test and will tell you if that test passed or if it failed. Also if it fails it will show the lines that diff.

The given test (tree_sum) has a wrong test case(test-2) just to show how the script handles the diff.

Now it only have two flags:

<table>
<tr>
	<td> Flag
	<td> Verbose
	<td> Description
<tr>
	<td> <code> -b
	<td> <code> --build
	<td> Build the hvm executable using <code> cargo build
<tr>
	<td> <code> -c
	<td> <code> --clean
	<td> Remove the target directory (<code>cargo clean</code>)
</table>


AdrianoII:
With the @quleuber help, we refactor the script, and now it tests the compiled version of each test case.
Besides that, I split our CI processes into "subworkflows" so now we have the Cargo workflow, Nix workflow, and the new Tests workflow.
The CI workflow just joins them and set the dependency of each workflow.
The Tests workflow runs our black-box tests if the cargo workflow succeeds.

AdrianoII:
Right now, I disabled the compiled tests on Windows because our runtime is not portable. But I am already looking into how to solve this.

quleuber (approved):

PR #114, opened 2022-05-03
fix #99 - flag to set quantity of allocated memory.

### LINKED ISSUE #99: (temporary) flag to set quantity of allocated memory
e.g.
```sh
hvm compile -M 2G main.hvm
```

samueldurantes:
This PR adds a flag that allows set initial memory allocation, now users with low ram machines can set a lower value to be allocated.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,5 @@
 mod builder;
+mod cli;
 mod compiler;
 mod language;
 mod parser;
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -70,19 +73,26 @@ fn run_cli() -> Result<(), String> {
...
       let code = load_file_code(&hvm(&file))?;
 
-      run_code(&code, true, params)?;
+      run_code(&code, true, params, parse_gb(&cli_matches.memory))?;
       Ok(())
     }
   }
 }
 
+fn parse_gb(raw: &str) -> usize {
```
quleuber, line 83: How hard would be for us to interpret units?

quleuber: If we are going to only support GB values we could drop the suffix altogether.

Also, `clap` seems to have support for custom parsing/validation (which would give better error messages).

quleuber:
Also, we are still not setting the memory size on the compiled version.

src/main.rs
quleuber: https://github.com/clap-rs/clap/blob/v3.1.15/examples/tutorial_derive/04_02_validate.rs

samueldurantes: > How hard would be for us to interpret units?

How many units do you think interesting?

samueldurantes: How many units do you think interesting? Particularly I think that MB and GB would be sufficient.

quleuber: I agreed that it would be sufficient. 
But I think if you just associate `K`, `M`, and `G` with the exponents (a list would suffice) it would be already more elegant and generic.
I think we should only use the prefix `M` instead of  `MB`, that also simplifies it.

samueldurantes: I think that would be good use the pattern of the International System of Units.

quleuber: The units should be [binary](https://en.wikipedia.org/wiki/Binary_prefix) actually.

PR #125, opened 2022-06-04
Better strings

Z-snails:
Strings are now converted to linked lists at runtime, not compile time
This prevents stack overflows in files with long strings
Also readback strings as actual strings runtime.c


VictorTaelin:
The idea is good but this PR breaks some conventions. I'll consider it later. Thanks for the contrib

PR #134, opened 2022-08-02
Add no-std support.

FranchuFranchu:
HVM only uses OS services for some things, including opening files and for the CLI program. Thus, a restricted subset of the HVM library can technically run without them. So I added a `std` feature, enabled by default, that compiles HVM to use the standard library. When it's disabled using `default-features = false` by a library that uses it, HVM will compile in `#![no_std]` mode, which prevents using the `std` library but allows running in embedded and kernel environments. 

I had to change some things to allow this. First of all, `regex` requires std, so the compiler to C is disabled in `no_std` mode. `std::collections::HashMap` was replaced by `hashbrown::HashMap` because std's requires the OS RNG. All imports from STD were changed with imports from `core` or `alloc`. Future commits should also do this to prevent new HVM versions from not compiling in `no_std`.

Performance should not be affected negatively. Hashbrown's benchmarks show that it's usually faster than std's hashmap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,17 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
+dependencies = [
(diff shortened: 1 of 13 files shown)


VictorTaelin:
Very cool, thank you. The only drawback is this will complicate slightly future updates, since we now need to think about what is on std and what isn't, right?

PR #144, opened 2022-09-09
fix reserved ids

racs4:
Maybe we should do something like
```rust
...
pub const RESERVED_QUANTITY: u64 = 4;



pub const FIRST_IO: u64 = RESERVED_QUANTITY + 1;
...
pub const IO_QUANTITY: u64 = 4;



pub const HOAS_CT0 : u64 = RESERVED_QUANTITY + IO_QUANTITY + 1;
...
```
?


VictorTaelin:
agreed

PR #145, opened 2022-09-12
Do arithmetic mod-2^32 in `quick_sort` test

### LINKED ISSUE #142: `quick_sort.hvm` test broken
[tests/quick_sort/quick_sort.hvm](https://github.com/Kindelia/HVM/blob/4345d5baf9b49b3ca59dfbcef00199baebcb038a/tests/quick_sort/quick_sort.hvm) hangs, but [bench/QuickSort/main.hvm](https://github.com/Kindelia/HVM/blob/4345d5baf9b49b3ca59dfbcef00199baebcb038a/bench/QuickSort/main.hvm) runs fine.

This is surprising, since they are almost identical -- the main difference is that `bench/` version has a `%` that is gone in the `tests/` version:

```diff
--- a/bench/QuickSort/main.hvm
+++ b/tests/quick_sort/quick_sort.hvm
@@ -10,7 +10,7 @@
 
 // Generates a random list
 (Randoms s 0) = (Nil)
-(Randoms s l) = (Cons s (Randoms (% (+ (* s 1664525) 1013904223) 4294967296) (- l 1)))
+(Randoms s l) = (Cons s (Randoms (+ (* s 1664525) 1013904223) (- l 1)))
 
 // Sums all elements in a concatenation tree
 (Sum Empty)        = 0
@@ -43,5 +43,5 @@ Pivot = 2147483648
 
 // Sorts and sums n random numbers
 (Main n) =
-  let list = (Randoms 1 (* 100000 n))
+  let list = (Randoms 1 (* 100 n))
   (Sum (QSort Pivot Pivot list))
```

The problem seems to be that:

* without the `%`, `Randoms` generates 60-bit integers
* but the initial pivot is hard-coded to 2^31
* so almost all elements fall to the right of the pivot
* so running time becomes `O (n * n)` rather than `O (n log n)`

I see two easy fixes:

* put back the `%`
* Use a 60-bit-friendly pivot (e.g. 576460752303423488)

lawcho:
Fixes #142

The test now calculates the expected answer, and quickly.

### DIFF
--- a/tests/quick_sort/quick_sort.hvm
+++ b/tests/quick_sort/quick_sort.hvm
@@ -10,12 +10,12 @@
 
 // Generates a random list
 (Randoms s 0) = (Nil)
-(Randoms s l) = (Cons s (Randoms (+ (* s 1664525) 1013904223) (- l 1)))
+(Randoms s l) = (Cons s (Randoms (% (+ (* s 1664525) 1013904223) 4294967296) (- l 1)))
 
 // Sums all elements in a concatenation tree
 (Sum Empty)        = 0
 (Sum (Single a))   = a
-(Sum (Concat a b)) = (+ (Sum a) (Sum b))
+(Sum (Concat a b)) = (% (+ (Sum a) (Sum b)) 4294967296)
 
 // The initial pivot
 Pivot = 2147483648


lawcho:
Note that the "benchmark version" of this test ([bench/QuickSort/main.hvm](https://github.com/Kindelia/HVM/blob/4345d5baf9b49b3ca59dfbcef00199baebcb038a/bench/QuickSort/main.hvm)) still contains the second bug (`Sum` computes the wrong answer).

VictorTaelin:
ty

PR #148, opened 2022-09-14
Platform IO

lawcho:
I'd like to see [Roc-style platform IO](https://www.youtube.com/watch?v=cpQwtwVKAfU&feature=youtu.be) in HVM (esp. for compiled programs), which has already been suggested [several](https://github.com/Kindelia/HVM/issues/106#issuecomment-1117243523) [times](https://github.com/Kindelia/HVM/issues/38#issuecomment-1099276995).

Here's what I think needs to be done:

- [x] Naively split runtime.c
- [x] Solve the constructor-pinning problem
- [x] Implement a ~line-based~ character-based IO platform
- [x] Split platform/runtime features
- [x] API design
- [x] Update old tests (use `no_io` platform)
- [x] Write examples for other platforms
- [x] Convert examples to automated tests
- [x] Debug failing tests
~- [ ] Write documentation~ (not in this PR, see comment)
~- [ ] Update interpreter~ (not in this PR, see comment)

Notes:

* You chose a platform at compile-time, e.g.
   `hvm c examples/echo.hvm -P platforms/console.c`
  The default platform implements the previous behavior (print normal form + stats)
* The constructor-pinning problem:
  * IO platforms may need to know the ids of certain constructors (e.g. `IO.do_output`)
  * But the constructor <-> id mapping is currently generated *per program*
  * ~Possible solution: at compile time, pass list of *n* constructors to pin to first *n* ids~
  * *Current plan: append platform to runtime, so it can use generated `#define`s*
* Platform/runtime feature split
  * Before, the runtime implemented everything
  * After, it may be more portable to make some features platform-specific (e.g. normal form printing)
  * There are several fine points to be decided here:
    * Which features stay in the runtime / get bumped to platforms?
    *Current plan: target-independent utilities (e.g. evaluator, read-back) go into main, everything else either
    goes into platforms (e.g. main, file descriptor handling), or can be disabled by a flag (e.g. pthreads)* 
    * Who combines the platform + runtime (hvm / user)?
      *Current plan: hvm (so building simple programs stays simple)*
    * How are the platform + runtime combined (concatenation / `#include` header-file)?
      *Current plan: concatenation (so that generated `#define`s are visible in platform)*
    * ~What API will the *runtime* expose to platforms?~
    * ~How do we keep generated code out of the API?~
* Differences to interpreter IO:
   * The interpreter's IO interface is big -- none of the platforms fully implement it
   * The interpreter's console IO is line-oriented, whereas `console.c` is character-oriented

I've made a start in this branch, but am new to the codebase (and to C programming), so **feedback is very welcome**


lawcho:
I'm leaving the last two tasks as future work, because

* updating the interpreter is a big job (we may need to JIT-compile platforms)
* documentation for platforms is about as complete as the rest of the codebase
  (there doesn't seem to be a a 'user manual' or similar place to put detailed user-facing documentation)

rigille:
this is great @lawcho!

steinerkelvin:
Oh lord, this is nice. Wish I could help with it rn

steinerkelvin:
@lawcho btw, I forgot to mention.

@VictorTaelin is working on porting everything to Rust on the `V2` branch.

lawcho:
Thanks for the heads-up @steinerkelvin -- I'll try to write platform code with porting to Rust in mind.

lawcho:
I'm closing this PR because:

* It's become big & hard to review
* I have some ideas for a simpler design
* It will need re-implementing in Rust soon

VictorTaelin:
Yes, sorry, this will become easier on the next version, we'll get rid of the C runtime. It would be insane to maintain IO on both back-ends. That version will include major simplifications and code quality improvements across the codebase, as well as bugfixes and much greater stability. It will make contributions much easier and allow HVM to grow into many directions.

lawcho:
@VictorTaelin what is the plan for IO in the Rust rewrite?

I'd like to use the generated runtime as an evaluator only, and implement my own domain-specific IO functions.

This approach has worked quite well for me so far (spread across this PR, and https://github.com/lawcho/rpphp).

Will a similar setup be possible with the new Rust-generating backend?

PR #158, opened 2022-10-11
Replace generated CIDs with LOOKUP_CID macro

lawcho:
# The problem

Currently, the compiler backed generates a C macro for each function name, e.g.

```c

//GENERATED_CONSTRUCTOR_IDS_START//
#define _GEN_ (31)
#define _SUM_ (28)
#define _MAIN_ (32)

//GENERATED_CONSTRUCTOR_IDS_END//
```
These macros are then used in generated case splits, e.g.
```c
         switch (fun)
          //GENERATED_REWRITE_RULES_STEP_0_START//
          {
            case _GEN_: { //<snip>
            };
            case _SUM_: { //<snip>
            };
            case _MAIN_: {//<snip>
            };

```

This causes the following issues:

* Encoding names with special characters (e.g. `IO.put_char`) requires [hard-to-get-right escaping code](https://github.com/Kindelia/HVM/blob/0bcbae286f995bc14895b0774914b1e19cf5e34e/src/compiler.rs#L32-L39) in the HVM compiler
* Downstream users of HVM may need to [re-implement this escaping code](https://github.com/lawcho/rpphp/blob/14ed49dc2e1f7a7de5e53a8c3f10107a1bfcd8da/platgen.hs#L63-L68)
* Even if the escaping is clash-free, the macro names can still clash with macros from libraries

# A solution

This PR fixes the issues above by replacing the family of `_COMPILED_NAME_` macros with a single `LOOKUP_CID` function macro, e.g.

```c
#define LOOKUP_CID(str) (\
(strcmp(str,"Gen")==0) ? 28 : \
(strcmp(str,"Main")==0) ? 32 : \
(strcmp(str,"Sum")==0) ? 31 : \
\
-1)
```
This macro is used similarly to before:
```c
          //GENERATED_REWRITE_RULES_STEP_0_START//
            if (fun == LOOKUP_CID("Gen") ) { //<snip>
              break;
            }
            if (fun == LOOKUP_CID("Main") ) { //<snip>
              break;
            }
            if (fun == LOOKUP_CID("Sum") ) { //<snip>
              break;
            }
```

# Benchmarks

In theory, this change increases the complexity of constructor-comparison from `O(1)` to `O(# constructors)`.
In practice, I did not see a noticeable slowdown on the test suite.

I think this is because the C optimizer evaluates `LOOKUP_CID(str)` fully at compile-time, when `str` is constant.

## Timings before this PR
```bash
time python3 tests/test_cli.py --run-mode compiled --hvm-cmd="hvm" >/dev/null

real    0m6,260s
user    0m4,669s
sys     0m1,645s
```
## Timings after this PR
```bash
$ time python3 tests/test_cli.py --run-mode compiled --hvm-cmd="./target/debug/hvm" >/dev/null

real    0m7,367s
user    0m5,424s
sys     0m1,613s

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -326,7 +326,7 @@ checksum = "c4a1e36c821dbe04574f602848a19f742f4fb3c98d40449f11bcad18d6b17421"
 
 [[package]]
 name = "hvm"
-version = "0.1.88"
+version = "0.1.89"
 dependencies = [
  "clap",
  "highlight_error",
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -29,15 +29,6 @@ fn compile_code(code: &str, heap_size: usize, parallel: bool) -> Result<String, 
   Ok(compile_book(&book, heap_size, parallel))
 }
 
-fn compile_name(name: &str) -> String {
-  // TODO: this can still cause some name collisions.
-  // Note: avoiding the use of `$` because it is not an actually valid
-  // identifier character in C.
-  let name = name.replace('_', "__");
-  let name = name.replace('.', "_");
-  format!("_{}_", name.to_uppercase())
-}
-
 fn compile_book(comp: &rb::RuleBook, heap_size: usize, parallel: bool) -> String {
   let mut c_ids = String::new();
   let mut inits = String::new();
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by lawcho** [comment]
Update: I've replaced the Rust-unfriendly C macro with an inline function, which gives similar performance.

I haven't seen an API for the Rust rewrite yet, but I'm guessing it will contain a function like `lookup_cid`.
If it does, then this PR brings the C runtime one step closer to the Rust version.

## Timings after this PR (commit 5d4b191)

```bash
$ time python3 tests/test_cli.py --run-mode compiled --hvm-cmd="./target/debug/hvm" >/dev/null

real	0m6,385s
user	0m4,986s
sys	0m1,344s

```

VictorTaelin:
Is this still relevant in the new version?

lawcho:
No, this PR only touches the C backend.

PR #159, opened 2022-10-22
Fix CI

samueldurantes:
This PR removes `Cargo Fmt` and `Cargo Clippy`, given that HVM has your own code style.


steinerkelvin:
Yeah, the fmt should be disabled.

But, regarding clippy, I believe we should globally disable stuff like the `clippy::style` group — perhaps some other lints contextually — and strive for fixing the lints.

PR #175, opened 2022-11-22
Add exit(1) if a command failed

Janiczek:
I'm not a Rust dev but this did the trick 😁 
This helped me chain commands in Bash scripts, eg. `hvm run -f "${MAIN_HVM}" && echo "Finished"` etc.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -66,6 +66,7 @@ enum Command {
 fn main() {
   if let Err(err) = run_cli() {
     eprintln!("{}", err);
+    std::process::exit(1);
   };
 }
 


zicklag (approved):

VictorTaelin:
Doesn't panic do that? I believe we had a better handle of that before. Should definitely improve this.

zicklag:
A panic will exit with code 101 ( I'm not sure if that code is stable or not ).

But if we run into an error, I don't think we want to panic. Panic should IMO be reserved for critical programmer errors that represent bugs that will need to be fixed, while cleanly exiting non-zero would indicate that there was an error performing the operation, but not a program bug.

VictorTaelin:
Yes I agree, there are many functions that could easily be converted to return Result. Should be striaght-forward.

zicklag:
[Anyhow](https://docs.rs/anyhow/latest/anyhow/) can be extremely useful for this kind of thing. Just have all functions return `anyhow::Result` and then you can use the `?` to unwrap any `Result` with an `Err` variant that implements `std::error::Error`.

That works great for CLI error propagation, and [`thiserror`](https://docs.rs/thiserror) is perfect for deriving error types for library/APIs.

Issue #179, opened 2022-11-24
Performance regression on HVM v1.0


developedby:
Doing some basic tasks with the new version of hvm, I'm seeing some considerable performance regression compared to the old version (v0.1.89)

With this simple fibonacci program:
```
(Fib 0 x0 x1) = x0
(Fib idx x0 x1) = (Fib (- idx 1) x1 (+ x0 x1))
(Main) = (Fib <n> 0 1)
```
The number of rewrites scales linearly with `n`, and so does the number of accumulated dup nodes.
On the old HVM, the time used for this program scales linearly. On the new HVM it scales quadractically, even though the number of rewrites and space used is the same on both..

If I make the `x1` variable strict by adding an extra rule like so, we don't accumulate dups anymore
```
(Fib 0 x0 x1) = x0
(Fib idx x0 x1) = (Fib (- idx 1) x1 (+ x0 x1))
(Fib idx x0 0) = 0
(Main) = (Fib 1000000 0 1)
```
The performance on the old HVM is the same as before, but on the new version it becomes linear.

Besides this slowdown from dups, I'm also seeing a large regression on the single-core interpreted execution of this same program. On the old HVM I get ~55MR/s, while on the new version I get just ~30MR/s.

I know @felipegchi also noticed a significant performance regression for running the Kind typechecker, seing a very large slowdown.

VictorTaelin:
Can you help me identifying the source of this regression? 

Regarding the constant-time slowdown, I expect the use of atomic u64s instead of u64 to cause such regression. The locking of dups could also add up. Moreover, the use of a Visit Stack and a Redex Bag could impact it. These things are necessary for synchronization. If that's the case, we could disable atomics entirely on single core, and aim to reduce the usage of atomics and locks to only when it is necessary on multiple cores. That would probably improve the peak parallel speed too.

Regarding a program taking quadratic time to perform the same number of rewrites, that makes absolutely no sense to me. Perhaps it is a bug where the `reduce` function returns prematurely for some reason? This would make the program reduce partially, which would cause `normalize` to execute repeatedly, thus re-traversing the term unnecessarily. That's the only thing I can think of. Would be good to debug and have a confirmation.

VictorTaelin:
The commit above fixes that issue, but it has some problems that must be fixed before we merge. Locks use too much space. I'll work on this later on. Can you test how that branch performs?

developedby:
[...]

PR #181, opened 2022-12-05
fix reduce

NaoEhSavio:
Previously this code did not run in version 1.0.5-beta.

(Fib 0 x0 x1) = x0
(Fib i x0 x1) = (Fib (- i 1) x1 (+ x0 x1))

(Main n) = (Fib 500000 0 1)
and this one in version greater than 1.0.5-beta.

(Head (Cons x xs)) = (Some x)
(Head Nil) = None

(UnwrapOr (Some x) y) = x
(UnwrapOr None y) = y

(SumFirstTwo Nil) = 0 
(SumFirstTwo (Cons x xs)) = (+ x (UnwrapOr (Head xs) 0))

(Iter f x 0) = x
(Iter f x n) = (f (Iter f x (- n 1)))

(Fib n) = (Head (Iter λxs (Cons (SumFirstTwo xs) xs) (Cons 0 (Cons 1 Nil)) n))

(Main n) = (Fib 10000)
Now with these changes both programs are working perfectly, without loss of performance...

### DIFF
--- a/src/runtime/base/reducer.rs
+++ b/src/runtime/base/reducer.rs
@@ -122,7 +122,7 @@ pub fn reducer(
             DP0 | DP1 => {
               match acquire_lock(heap, tid, term) {
                 Err(locker_tid) => {
-                  break 'work;
+                  continue 'work;
                 }
                 Ok(_) => {
                   // If the term changed, release lock and try again


VictorTaelin:
Thanks for your help!

To explain what is going on here, what @NaoEhSavio's commit accomplishes is: it removes the delay vector, which was used previously as follows. When a thread reached a dup node that was locked, it would add that location to the delay vector, and would move on to do other things. The problem is, once free, this thread would start locking dups itself, which, under the wrong conditions, would cause active threads to bump into locked dups endlessly, popping from their `.delay` vectors repeatedly, essentially doing long useless cycles/walks through subgraphs. This caused them to take quadratic time to compute a program, in function of the total number of rewrites - which violates HVM's complexity model. That problem was first reported on #179, and can be seen in action here: https://gist.github.com/VictorTaelin/ffae77756afcf463b7c52a08c9d6d98e. Notice how, after a while, threads 5 and 1 will walk through long needless loops to find their redexes.

In order to solve this issue, I attempted to replace the `delay` object by annotating locks with the continuation instead. That way, when a thread hits a locked dup, it just leaves a note on it, and then moves on to other things. Then, when the dup is reduced, it moves the continuation to the thread that reduced it. This solves the issue above, since a dup reduction will immediately trigger the continuation of its bound variables. The problem is, in some cases, including the infamous program below, threads would all end up annotating their continuations in some dup, and move to steal mode, until a point where every thread is stealing and no thread is doing any work. This state shouldn't be reached, because it will cause progress to cease.

```
// List type
(Head (Cons x xs)) = (Some x)
(Head Nil) = None

// Something like Rust's Option.unwrap_or
(UnwrapOr (Some x) y) = x
(UnwrapOr None y) = y

(SumFirstTwo Nil) = 0 
(SumFirstTwo (Cons x xs)) = (+ x (UnwrapOr (Head xs) 0))

// Iterate a function n times on an input
(Iter f x 0) = x
(Iter f x n) = (f (Iter f x (- n 1)))

(Fib n) = (Head (Iter λxs (Cons (SumFirstTwo xs) xs) (Cons 0 (Cons 1 Nil)) n))

(Main n) = (Fib 20000)
```

As such, I rolled back the solution above and instead attempted to fix the redundant paths, while keeping the `delay` object. My plan was to keep it, but have a lightweight notification system, which would allow a thread that reduced a `dup` node to signal to the thread that has the respective continuation on its delay velay. This would cause that thread to prioritize the freshly-reduced location when popping from the delay vector, cutting the long/useless pathing short. I didn't like the added complexity, but decided to try implementing it. Midway through the implementation, @NaoEhSavio told me that he has been messing with the code and realized that all the cases we've tested work if we just change `break 'work` to `continue 'work`.

What this causes is that we remove the `delay` object entirely. Instead, when a thread arrives in a locked dup, it will just stay there waiting for it to be reduced forever. Interestingly, and contradicting my own memory, this solution doesn't seem to cause issues for the program above. Instead, it indeed just works for all cases. As such, I'm accepting this PR **for now**, as we better understand the problem.

I believe this solution won't work on the long term because it causes a thread to depend on the reduction of a dup node to continue. As such, the thread it is waiting for ends up arriving at a dup node that this thread locked, we'll have a dead-lock and the program will freeze. But, now, I'm not sure this actually can happen in normal programs, although intuitively it should, and I believed the `fib` example above was a case where this occurred. So, again, for now, I'll accept this solution, until we recall a program where such deadlock occurs. Once (if) that does happen, we'll need to come up with another solution to the deadlock problem, that doesn't involve the delay object, which is problematic for the case shown in #179.

PR #191, opened 2023-01-12
Added an example file for 1D Matrix Multiplication

ksutaria:
Created this example to test the limits of what has been built. Unfortunately, I get a stack overflow error when multiplying  100K+ values in a 1D list with itself. Any thoughts on how to account for this stack overflow scenario or criticism on coding style welcome. Thanks!

### DIFF
--- a/examples/matrices/matrix-multiplication.hvm
+++ b/examples/matrices/matrix-multiplication.hvm
@@ -1,0 +1,45 @@
+// Generates a random list
+(Rnd 0 s) = Nil
+(Rnd n s) = (Cons s (Rnd (- n 1) (% (+ (* s 1664525) 1013904223) 4294967296)))
+
+// Scale a list by 's'
+(Scale s Nil) = (0)
+(Scale s (Cons t Nil)) = (Cons (* s t) Nil)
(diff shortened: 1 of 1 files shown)


ksutaria:
Another issue that I noticed was that certain versions of main would go through proper reduction while others don't. Any insight into what I may be missing is helpful. Thanks!

VictorTaelin:
Thank you so much. I'll have a look at the issues when I'm back to coding :)

mgubi:
You definitions of `Scale` and `Multiply` do not seems to be correct (for example `Scale` does not return a list in all the cases, or `Multiply` does not recurse on the length of the lists).  Alternatives are:
```
// Scale a list by 's'
(Scale s Nil) = Nil
(Scale s (Cons t l)) = (Cons (* s t) (Scale s l))

// Multiply two 1D lists
(Multiply a Nil) = Nil
(Multiply Nil a) = Nil
(Multiply (Cons a b) (Cons c d)) = (Cons (* a c) (Multiply b d))
```
and they run fine for me.  I'm not sure I understand what you want to do with `Multiply`, my version just take the product of two lists element-wise to produce a new list.

PR #193, opened 2023-01-18
fix: fix `Runtime::new` and `Runtime::from_code_with`.

zicklag:
Runtime creation was either panicking or getting killed by the system because it was creating as many threads as the heap size, instead of reading the thread count variable.


VictorTaelin:
Thanks!

PR #195, opened 2023-01-19
refactor: apply all clippy lints and cargo format.

zicklag:
Not sure if you guys are ready for this or not, but here it is if it helps!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -217,7 +217,7 @@ checksum = "809e18805660d7b6b2e2b9f316a5099521b5998d5cba4dda11b5157a21aaef03"
 
 [[package]]
 name = "hvm"
-version = "1.0.20-beta"
+version = "1.0.0"
 dependencies = [
  "backtrace",
  "clap",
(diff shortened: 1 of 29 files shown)


zicklag:
I also switched the rust toolchain to nightly in CI, because the crate doesn't compile on stable right now, but it doesn't seem like the CI jobs will run anymore.

zicklag:
Actually, I just realized it'd be really easy to support stable. We'd just have to polyfill the two nightly methods we're using I think.

I could try to add that.

VictorTaelin:
Thank you so much! I really want to accept this commit but there are many things that similar PRs broken before (like some alignments here and there) so I'll need to have a look on everything before I'm able to merge it. I'll be back to coding when I'm done with fundraising work. Great work, thanks

zicklag:
Sounds good!

To ease the review burden, as I know this is a ton of changed code, maybe I should do separate PRs for the automatic application of `cargo fmt` and the clippy lints which were semi-automatic and semi-manual.

Once I get the chance, I'll try to open 3 PRs:

- Cargo format
- Clippy fix
- Polyfill nightly methods so we can compile on stable.

PR #197, opened 2023-01-24
chore: update Cargo.lock.

zicklag:
The lockfile was not up-to-date with the current crate version.

As soon as you run `cargo build` it updates the Cargo.lock to match, but it'd be good to have it in sync with the current Cargo.toml.


zicklag:
superseded by #200.

PR #200, opened 2023-01-24
fix: add missing `proptest` library to dev dependencies.

zicklag:
The `proptest` crate is used in the tests, but wasn't added to the dev dependencies.

This also updates the Cargo.lock file to match the current repo version ( i.e. fixes #197 ).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,3 +27,6 @@ crossbeam = "0.8.2"
 highlight_error = "0.1.1"
 instant = { version = "0.1", features = [ "wasm-bindgen", "inaccurate" ] }
 itertools = "0.10"
+
+[dev-dependencies]
+proptest = "1.0.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,6 +50,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
+dependencies = [
+ "bit-vec",
+]
+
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -27,3 +27,6 @@ crossbeam = "0.8.2"
 highlight_error = "0.1.1"
 instant = { version = "0.1", features = [ "wasm-bindgen", "inaccurate" ] }
 itertools = "0.10"
+
+[dev-dependencies]
+proptest = "1.0.0"
```
nightscape, line 32: It's generally a good idea to put a newline at the end of every text file:
https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file
I usually recommend to configure your editor to do it automatically (almost all editors can).

zicklag:
[This test](https://github.com/Kindelia/HVM/blob/5c25f8aefc777029816698e2dc476a2646380726/src/language/rulebook.rs#L617) is the only one failing. I think it may be related to a change in the way programs are parsed, not an actual bug in the code, but I'm extremely unfamiliar with the codebase so I'm not sure.

```
---- language::rulebook::tests::test_rulebook_expected stdout ----
thread 'language::rulebook::tests::test_rulebook_expected' panicked at 'assertion failed: `(left == right)`
  left: `33`,
 right: `3`', src/language/rulebook.rs:638:5
```

Cargo.toml
zicklag: 👍 Updated.

nivkner:
the reason for the failure is that precompiled functions are added to the rule book, presumably so their symbols could be resolved in user code.

https://github.com/HigherOrderCO/HVM/blob/ba9062de86cf8ff7fba1283f88b467fb21db13eb/src/language/rulebook.rs#L37-L43

PR #201, opened 2023-01-24
style: manually fix remaining clippy lints.

zicklag:
This depends on #199  to avoid conflicts, and therefore contains the commit from https://github.com/Kindelia/HVM/pull/198 in it. I can rebase this when #199 is merged so that only one is a part of the PR.

---

This manually applies all clippy lints that can't be applied automatically.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -10,7 +10,6 @@ pub fn eval(
   tids: usize,
   dbug: bool,
 ) -> Result<(String, u64, u64), String> {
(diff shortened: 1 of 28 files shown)


zicklag:
Thanks for catching that! I'm not sure when this will get reviewed, and it looks like there are quite a few conflicts now, so I'm going to wait to make any changes until I get some confirmation that this is in the right direction.

Issue #188, opened 2022-12-21
Compiled HVM programs need a better CLI interface


developedby:
Running a program compiled with `hvm compile somefile.hvm` and then `cargo build --release`, I am greeted with this message

```
$ ./target/release/somefile 

somefile 1.0.17-beta
A massively parallel functional runtime.

USAGE:
    somefile <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    compile    Compile a file to Rust
    help       Print this message or the help of the given subcommand(s)
    run        Load a file and run an expression
```

That means that this program is not carrying only the interpreted and precompiled functions, but also the whole compiler.

That is unintuitive, what I expected was to just be able to call my program directly (which would correspond to calling it with `somefile run "(Main arg)"`).

We should probably create a separate CLI for the compiled programs to make this simpler to use

VictorTaelin:
Makes sense to me. Will add.

PR #202, opened 2023-01-24
Add REPL, improve CLI, split project, update Nix files

nothingnesses:
* Split the project into a library crate "hvm" and executable crate "hvm-cli", under one workspace. This should address [this comment](https://github.com/HigherOrderCO/HVM/pull/239#issuecomment-1518652141).
* Made a separate and simpler template file for the compiled outputs which should address #188.
* Made a simple REPL which should address [this comment](https://github.com/Kindelia/HVM/issues/91#issuecomment-1066745685).
* Updated the Nix files.

In the future, it might be nicer if HVM is able to read from a TOML file (or some other configuration format) to be able to set the contents of the resulting Cargo.toml file, especially its version and description fields.

As an aside: it might also be cool to have HVM also be able to output Rust libraries instead of just binaries, which would allow some form of FFI from Rust to HVM.


nothingnesses:
Oh wait, I think it'd be better if the shared functions in main.rs and compiled-cli.rs are refactored into a single module which both those files can import to prevent repetition. I'll work on that tomorrow and will update these commits after.

PR #203, opened 2023-01-25
refactor: polyfill nightly functions and switch to stable rust.

zicklag:
Depends on #201 to avoid merge conflicts. I'll rebase this after that is merged.

---

This makes hvm compile on stable Rust! The annoying part is that we must manually identify which rust platforms have their integer sizes exactly equal to their alignment. Since the `target_has_atomic_equal_alignment` cfg check is unstable, we have to create our own equivalent, which works by whitelisting platforms where we know that the `u8` and `u64` sizes are of the alignments `8` and `64`.

This is easy to validate by looking at the [data layout](https://llvm.org/docs/LangRef.html#data-layout) of the target ( see [here](https://users.rust-lang.org/t/is-there-a-way-to-emulate-atomic-alignment-cfg-check-on-stable/87863/3?u=zicklag) ), so it's not that big a deal, we will just have to whitelist all supported platforms that people need to build for.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,29 @@
+fn main() {
+  let target = std::env::var("TARGET").unwrap();
+  // This is is part of a larger workaround to support compiling on stable rust.
+  //
+  // This is the set of platforms that have been manually verified to have atomic integers with the
+  // same alignment as the normal integers. There are doubtless platforms that satisfy the
+  // requirement that aren't in the list, and should be added if anybody needs to compile for the
+  // platform. Pull requests welcome.
+  //
+  // This is a workaround for the unstable `cfg(target_has_atomic_equal_alignment = "n")` check, and
+  // should be removed once that or the `atomic_from_mut` and `atomic_mut_ptr` features are stable.
+  //
+  // See also `polyfills.rs`.
+  //
+  // See
+  // https://users.rust-lang.org/t/is-there-a-way-to-emulate-atomic-alignment-cfg-check-on-stable/87863/2
+  // for background on how to verify compatible platforms.
+  let targets_that_have_equal_alignments = &[
+    "wasm32-unknown-unknown",
+    "x86_64-unknown-linux-gnu",
+    "x86_64-unknown-linux-musl",
+    "x86_64-pc-windows-msvc",
+    "x86_64-pc-windows-gnu",
+    "x86_64-apple-darwin",
+  ];
+  if targets_that_have_equal_alignments.contains(&target.as_str()) {
+    println!("cargo:rustc-cfg=stable_target_has_atomic_equal_alignment");
+  }
+}
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly"
+channel = "stable"
--- a/src/api.rs
+++ b/src/api.rs
@@ -10,7 +10,6 @@ pub fn eval(
   tids: usize,
   dbug: bool,
 ) -> Result<(String, u64, u64), String> {
-
   // Parses and reads the input file
   let file = language::syntax::read_file(&format!("{}\nHVM_MAIN_CALL = {}", file, term))?;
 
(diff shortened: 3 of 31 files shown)


zicklag:
After merging #198, #199, #201, #203 ( this PR ), #200, and fixing the one failing test, CI should be totally green!

nivkner:
code generated with `hvm compile` fails to compile because `build.rs` and `src/polyfills.rs` are not copied to the output crate in `src/compiler/mod.rs`

nivkner:
having looked more into the code i came to the conclusion that the functions being polyfilled here are almost entirely unnecessary.
https://github.com/HigherOrderCO/HVM/blob/ba9062de86cf8ff7fba1283f88b467fb21db13eb/src/runtime/base/memory.rs#L390
could be replaced with
```rust
(0..size).map(|_| AtomicU8::new(0xFF)).collect()
```
this is actually identical to the unsafe and unstable version once optimized, a single allocation and a `memset`, see [the godbolt link](https://godbolt.org/z/3Tsqaxe4c)

`as_ptr` is used primarily for thread local data (`heap.lvar`, `heap.aloc`, `heap.vbuf`), which is accessible via the `Heap` but is actually only ever accessed by the single thread with the corresponding `tid`.

the simplest way to implement this on stable is to wrap each such data structure is an `UnsafeCell`, and use `UnsafeCell::get` when accessing it. so long as only the thread assigned a given `tid` accesses the element at index `tid`, this is safe. this is also the invariant needed for the current implementation so no furthur restrictions.

the calls to `get_cost`, which accesses the cost of all threads, are done in between reductions where you have exclusive access to the heap, so you can use `UnsafeCell::get_mut`. `get_used` is unused.

the safer way is to actually make them only accessible to a single thread, either by a thread local, or by passing a thread local handle to the heap as follows:

```rust
pub struct HeapHandle<'a> {
  pub tids: usize,
  pub node: &'a [AtomicU64],
  pub lock: &'a [AtomicU8],
  pub lvar: LocalVars,
  pub vstk: &'a [VisitQueue],
  pub aloc: Box<[AtomicU64]>,
  pub vbuf: Box<[AtomicU64]>,
  pub rbag: &'a RedexBag,
  pub global_cost: &'a AtomicU64,
}
```

then you add the thread local cost to the global cost once the handle is dropped at the end of each reduction.

the only sticking point is https://github.com/HigherOrderCO/HVM/blob/ba9062de86cf8ff7fba1283f88b467fb21db13eb/src/runtime/base/program.rs#L187
which relies on the fact that after allocation access to the newly allocated nodes is exclusive.
this can be handled by making reference types `NodeRef<'a>` and `NodeMutRef<'a>` which internally uses the tagged pointers, but grant immutable and mutable access to `heap.node` respectivly. then allocation could grant `NodeMutRef<'a>` while reading pointers from existing nodes will only get `NodeRef<'a>`. then its just a matter of using `AtomicU64::get_mut`.
(in this case you would actually need a slice version).

this also handily solves the usage of `get_unchecked` since the reference types are guaranteed to originate from within the heap.

a quick and easy solution would be to just do a relaxed atomic store here, this is what happens in the `link` above it anyway.

VictorTaelin:
Looks promising, but I still don't have the time to go through this change rn ):

nivkner:
thinking about `NodeMutRef` some more:

the current allocation strategy for nodes is to iterate over the nodes array, checking for the first open spot, and allocating there.
you can see that here:
https://github.com/HigherOrderCO/HVM/blob/ba9062de86cf8ff7fba1283f88b467fb21db13eb/src/runtime/base/memory.rs#L450-L457

this means that if you hand out a mutable reference (or something semantically similar), you cannot allocate more terms while holding on to it, since that would be undefined behavior.

while it might be possible to enforce this with lifetimes on the `NodeMutRef`, the result might get difficult to use.

some alternatives are:
1. use external information in allocator to determine available nodes like a free-list (could be in the free nodes like in `slab` crate) to avoid reading from allocated nodes.
2. only initialize during allocation either by passing the value (like `Box::new`) or by passing a closure which will be called at allocation.
3. only use `NodeRef` and stick with atomic accesses.

PR #206, opened 2023-01-26
Speed up bubble sort benchmark with extra bangs

edmundnoble:
This gives me a ~3x speedup or so.


VictorTaelin:
Thank you! To keep the comparisons faithful we must apply the same on the HVM side (which has `!` annotations, although not as good as Haskell's). Also, we must update the benchmarks to reflect the updated versions (otherwise the README would be outdated). I can work on that later on, will merge if someone does it.

VictorTaelin:
Thanks but we still need to regenerate the image on the README with the new benchmark. I'll ask someone

PR #231, opened 2023-04-03
Add HVM.seq and `seq` syntax.

FranchuFranchu:
This PR adds a `seq` syntax, similar to ٰ`ask` and `let`, to allow for sequential computation in HVM. It is useful for working with global variables. It is syntax sugar for a new precomp function `HVM.seq`, which can be implemented in pure HVM code but I've decided to make it builtin.

I've also added examples in the `examples` directory.

Example usage:
```
SideEffectful = (HVM.log "Banana" 0)

Main = 
        // Usecase one: Running functions with side-effects where you discard the value
        seq SideEffectful;
	// Usecase two: Working with global variables.
	let some_val = (@$value(20) 10)
	seq twenty = some_val;
	seq (HVM.log $value 0);
	twenty
```


VictorTaelin:
Thank you so much Franchu! I've left some comments on the code.

PR #239, opened 2023-04-18
use less dependencies (this time without requiring furthur changes)

nivkner:
this reverts the reversion of the commit that reduces the amount of dependencies needed when using HVM as a library.

since it would be unintuitive to require a `--feature ...` flag when installing HVM as a binary,  the default features include `"cli"`.
this also means that no furthur changes to the README are required. at least until more documentation is added which is geared towards using HVM as a library.

in general it would be cleaner if this project was split into 2 seperate creates `hvm` and `hvm_cli` and have the `hvm_cli` depend on `clap`, and only use the public API exported by `hvm`.

however that would require restructuring the code, which would be a problem so long as HVM compilation depends on the current structure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,21 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "addr2line"
-version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
-dependencies = [
- "gimli",
-]
-
-[[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
-
-[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


zicklag:
> in general it would be cleaner if this project was split into 2 seperate creates hvm and hvm_cli and have the hvm_cli depend on clap, and only use the public API exported by hvm.

I agree. In my own little experiments for other projects that's seemed to be the cleanest way to organize it whenever I have a CLI and an API that might be used without the CLI.

VictorTaelin:
Agree with the split. I'll be focusing on HVM soon(ish) and will be able to go through these PRs. Thanks (:

Boscop:
Btw, for the CLI flags like `--cost` and `--debug`, currently they require passing `true` (`-c true` / `-d true`):
https://github.com/HigherOrderCO/HVM/blob/ba9062de86cf8ff7fba1283f88b467fb21db13eb/src/main.rs#L40-L46

but that's not necessary: (https://docs.rs/clap/3.2.23/clap/_derive/_tutorial/index.html#flags)
```rs
    /// Shows the number of graph rewrites performed.
    #[clap(short, long)]
    cost: bool,

    /// Toggles debug mode, showing each reduction step.
    #[clap(short, long)]
    debug: bool,
```
and after upgrading to clap 4 it would be: (https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#flags)
```rs
    /// Shows the number of graph rewrites performed.
    #[arg(short, long)]
    cost: bool,

    /// Toggles debug mode, showing each reduction step.
    #[arg(short, long)]
    debug: bool,
```
It's a small change, it can be done in the process of factoring out the CLI into a `hvm-cli` crate & upgrading to clap 4..

Issue #232, opened 2023-04-08
fixed-point combinator hangs when using more than one thread


nivkner:
using the following code:

```idris
Y = λf ((λx (f (x x))) (λx (f (x x))))

Factorial = λf (λn ( U60.if (== n 0) 1 (* n (f (- n 1)))))

(Main n) = (((Y) Factorial) n)
```

the following command hangs:
`hvm run -f fixed_point.hvm "(Main 5)"`

whereas the following command terminates with `120` as expected:
`hvm run -f fixed_point.hvm "(Main 5)" --debug`

tested on commit `d3367849ec92a043ebf04fe537e9b74c516347b6`, with rust `nightly-2023-03-01`

Edit: might be a deadlock of some sort, which debugging prevents by serializing execution.
since similar results are gotten when specifying `-t 1` instead

Boscop:
I noticed the same behavior on my machine, also with the Y combinator:
```js
Y = λf ((λx (f (x x))) (λx (f (x x))))
(Main n) = ((Y) (λf λi (U60.if (== i n) Done (f (+ i 1)))) 0)
```
Without `--debug` flag, it works when calling `(Main 2)`, but with `(Main 3)` it uses all CPU cores and doesn't return..

Issue #183, opened 2022-12-05
Programs with a lot of dups get stuck


developedby:
```
(Func 0 b) = b
(Func a b) = (Func (- a 1) (+ b b))

(Main) = (Func 100000000 2)
```

Running this code with a very large `a`, makes it seemingly never end (expected 10s, waited 10m). Other programs that create lots and lots of dups also seem to cause this.

developedby:
Also it seems this game spends most of the time doing exception handling

VictorTaelin:
Is this for sure related to bugs? I see the program runs normally up to 1m, but performance downgrades quickly after ~4m. I'd guess that's caused by either heap, visit queue or redex bag being overload, so it will be improved only once we have growable memory. That's just a guess though, so I'd appreciate if you could measure the occupancy of each of these structures and report back.

developedby:
Not for sure related to dups strictly, but it does happen with them.

I'll make these measurements and report back

developedby:
Looks like the the visit queue is overflowing in this examples. When that happens, it actually continues storing and reading outside, sometimes at indexes 3x larger than the last position of the stack, before finally freezing or crashing.

Setting the heap to a small value to intentionally cause an overflow also results in similar behaviour.

So the problem here is not the dups, but rather that it accumulates a lot of `(+ b b)` to reduce at once (if I understood correctly what the visit stack is).

The redex bag never gets full. But I found it strange that it always increases its index. What happens if it grows too quickly, will it just overwrite an old unresolved value (and cause incorrect behaviour)?

Issue #60, opened 2022-02-03
Quadaric time & space on heavily shared lazy lists


oxalica:
Consider the program calculating `sum (map head (tails list))`. Due to the laziness, though the normal form of `tails list` has `O(n^2)` elements, I'm expecting the whole expression to cost linear time since we don't need to fully evaluate each lazy lists, and the number of elements we actually traversed are just `n`.
But unfortunately, HVM cost quadaric time & space on this scenario, indicating it's not lazy on sub-lists and all sub-lists elements are actually generated.

HVM:
```
(Gen 0) = Nil
(Gen n) = (Cons n (Gen (- n 1)))

(Foldr f z Nil) = z
(Foldr f z (Cons x xs)) = (f x (Foldr f z xs))

(Tails Nil) = (Cons Nil Nil)
(Tails (Cons x xs)) = (Cons xs (Tails xs))

(Map f Nil) = Nil
(Map f (Cons x xs)) = (Cons (f x) (Map f xs))

(Head Nil) = 0
(Head (Cons x xs)) = x

(Main n) = (Foldr @a@b(+ a b) 0 (Map @l(Head l) (Tails (Gen n))))
```

Haskell:
```haskell
import Prelude hiding (head, foldr, tails, map)
import Data.Word (Word32)
import System.Environment (getArgs)

data List a = Nil | Cons a (List a)

gen 0 = Nil
gen n = Cons n (gen (n - 1))

foldr f z Nil = z
foldr f z (Cons x xs) = f x (foldr f z xs)

tails Nil = Cons Nil Nil
tails (Cons x xs) = Cons xs (tails xs)

map f Nil = Nil
map f (Cons x xs) = Cons (f x) (map f xs)

head Nil = 0
head (Cons x xs) = x

main = do
    n <- read . (!! 0) <$> getArgs :: IO Word32
    print $ foldr (\a b -> a + b) 0 (map (\l -> head l) (tails (gen n)))
```

Run time:
| n | HVM (clang, O2, single thread) | GHC, O2 |
|-|-|-|
| 1000 |131ms | 10ms|
| 2000 | 316ms| 10ms|
| 4000 | 1082ms| 12ms|
| 8000 | 5.039s (1.6G mem)| 15ms (5.0M mem)|
|16000|31.018s (5.9G mem)|20ms (5.6M mem)|

[...]

Issue #44, opened 2022-02-01
Incorrect result on some inputs


Ekdohibs:
HVM produces incorrect results on some inputs, one of which being:
```
(Main arg) = ((λa (a (λb ((λc (λd (b (c d)))) a)))) (λe (e e)))
```
On this input, HVM produces the result `λx1 λx2 ((x1 x2) (x1 x2))` while the correct result is `λx1 λx2 ((x1 x1) (x2 x2))` (as verified by [https://crypto.stanford.edu/~blynn/lambda/](), with the corresponding input syntax by adding dots after the variable of each lambda: `((λa. (a (λb. ((λc. (λd. (b (c d)))) a)))) (λe. (e e)))`).

Besides, on other inputs such as:
```
((λa λb ((b a) λc λd (λe (d a) a)) λf (f f)) λz z)
```
HVM produces the result `λx1 (x1 λx2 (x2 _))` which is not even well-formed.

VictorTaelin:
This seems to be violating the "clones can't clone their own clones" limitation, explained [here](https://github.com/Kindelia/HVM/blob/master/HOW.md#superposed-duplication). HVM is just a (much) faster functional runtime based on the abstract algorithm, not a full λ-calculator. Attempting to make a clone clone its own clone is considered undefined behavior. I stress that this only happens in highly artificial examples (like these!).

L-as:
This limitation should probably be noted in the README. You probably also want to add more information on the limitation to make clear under what situations this is (im)possible. I assume this limitation is impossible to hit if the type system precludes applying a function to itself?

Ekdohibs:
While I agree it is ok if there are limitations, I find it a completely different matter to have limitations and to silently return an incorrect result!
Besides, it seems to me that this limitation could be easily violated in a "natural" program such as mapping over a list of lists: `map2 f l = map (map f) l`. The function `map` has to clone the function given to it (as it needs to call it several times), but `map f` contains a clone of `map` itself!

VictorTaelin:
[...]

PR #243, opened 2023-04-29
test HVM against a naive implementation of its rewrite rules

nivkner:
a good way to test if something is correct is to compare its outputs to a reference implementation which is easier to check for correctness manually.

this uses proptest, which HVM was already using, though was missing from the `Cargo.toml`, so i added it back, like in #200, except I also changed the tests so that they pass by taking into account the pre-compiled rules.

it heavily relies on #237 for directly manipulating `Term`s, instead of just evaluating strings. so it current contains it commits.

the naive implementation currently does not handle constructors, and consequently no user-generated rules, so proptest doesn't generate those. i skipped handling let expressions since the way it is used in HVM is basically syntactic sugar for application on a lambda, not sure why it isn't actually sugar.

i made sure to maintain linearity while normalizing terms and only undo it at the end, in case some of the rules depend on this property. didn't maintain globally unique names since all expressions are scoped in the AST, so there is no ambiguity we substituting.

this is a draft to get some general feedback on this approach, and to see if this is something you'd be interested in.


Boscop:
Yes, it makes a lot of sense to do this. I was also thinking that it makes sense to have a high-level/research variant of the HVM which is not optimized for performance but is instead optimized for 
1. readability / making it easy to reason about its code (and [formally verifying its correctness](https://github.com/HigherOrderCO/HVM/discussions/120))
2. having as little `unsafe` code (or micro-optimizations that reduce readability) as possible
3. quicker prototyping of different new ideas/features, such as new node types / new rewrite rules or static analysis / compile-time inet transformation passes (which is where a lot of performance gains can come from, with a proper IR)

And then having a test suite (with proptests & manual/regression tests) in place to ensure that the hand-optimized version of the HVM has the same semantics as the high-level/research version.
Also fuzz testing the HVM to find terms on which it doesn't behave as expected, e.g. to catch things like https://github.com/HigherOrderCO/HVM/issues/232 or https://github.com/HigherOrderCO/HVM/issues/183.
Also it could make sense having a test suite that compares HVM's output to a lambda calculus evaluator and compare their costs to catch things like https://github.com/HigherOrderCO/HVM/issues/60.
We could also use randomly generated terms to find terms where the lambda-calculus-reduced result differs from the interaction-net-reduced result, to get a clearer picture of the set of terms for which they differ (for addressing https://github.com/HigherOrderCO/HVM/issues/44).

PR #253, opened 2023-05-20
bump deps + linting

load1n9:


src/runtime/base/debug.rs
```diff
@@ -116,10 +116,8 @@ pub fn show_at(heap: &Heap, prog: &Program, host: u64, tlocs: &[AtomicU64]) -> S
     tlocs: &[AtomicU64],
   ) -> String {
     let term = load_ptr(heap, host);
-    let done;
```
howsiyu, line 119: It's better to have
```rust
let done = if term == 0 {
        "<>".to_owned()
    } else {
      match get_tag(term) {
```

src/runtime/rule/dup.rs
```diff
@@ -82,7 +109,7 @@ pub fn apply(ctx: ReduceCtx) -> bool {
   // x <- N
   // y <- N
   // ~
-  else if get_tag(arg0) == U60 {
+  if get_tag(arg0) == U60 {
```
howsiyu, line 112: What's the point of these changes? If you want to avoid triggering clippy::if_same_then_else you should just merge the U60 block and F60 block together. Better yet, use match instead of chaining if-else.

src/language/syntax.rs
```diff
@@ -300,19 +300,17 @@ pub fn parse_num(state: HOPA::State) -> HOPA::Answer<Option<Box<Term>>> {
...
       let (state, head) = HOPA::there_take_head(state)?;
-      Ok((state, ('0'..='9').contains(&head)))
+      Ok((state, head.is_ascii_digit()))
     }),
     Box::new(|state| {
       let (state, text) = HOPA::there_nonempty_name(state)?;
       if !text.is_empty() {
         if text.starts_with("0x") {
-          return Ok((state, Box::new(Term::U6O { numb: u60::new(u64::from_str_radix(&text[2..], 16).unwrap()) })));
+          Ok((state, Box::new(Term::U6O { numb: u60::new(u64::from_str_radix(&text[2..], 16).unwrap()) })))
```
howsiyu, line 309: I'm surprised you added #[warn(clippy::manual_strip)] but not used `text.strip_prefix("0x")`.

PR #270, opened 2024-05-08
NIX.md: update link to point to official wiki

Mic92:


kings177:
thank you, and sorry for the delay in the response, this repo now  is changed to HVM2 and we don't have a NIX set yet for the HVM2, however, [HVM1](https://github.com/HigherOrderCO/HVM/tree/hvm1) still exists, and i'll change make this change to the official Nix wiki there

PR #279, opened 2024-05-16
clean up, up until interaction table

enricozb:
clarifies syntax, allowing for flexibility of notation and identification of nodes and tress (non-var nodes).

adds a diagram, improved interaction rules, and their explanations.


kings177 (approved):
👍

PR #290, opened 2024-05-17
Update PAPER.typst

Felipegalind0:
fixed typo 
`and therefore possible to represent` 

by replacing with 
`and it is therefore possible to represent`

Final text:

```
This textual system is capable of representing any arbitrary Interaction Net, and it is therefore possible to represent "vicious circles" @Lafont_1997. We only consider HVM2 programs which do not contain any vicious circles.
```

fix typo `impactiful` to `impactful`


kings177 (approved):
thank you

PR #294, opened 2024-05-18
Make HVM available on Windows

sepcnt:
Get run-c and run-cu works on Windows 11. 
- `/experimental:c11atomics` and `-std=c11` is required for stdatomic
- Extremely large array (>=2GB) is not allowed on Windows so we have to use a pointer instead
- CUDA also needs some work for allocation
- threads.h rather than pthreads.h for compatibility.
- time64() with QueryPerformanceCounter()

### DIFF
--- a/src/hvm.c
+++ b/src/hvm.c
@@ -131,8 +131,8 @@ typedef struct Def {
   u32  node_len;
   u32  vars_len;
   Port root;
-  Pair rbag_buf[32];
-  Pair node_buf[32];
+  Pair rbag_buf[0xFFF];
+  Pair node_buf[0xFFF];
 } Def;
 
 // Book of Definitions
@@ -150,8 +150,8 @@ typedef struct TM {
   u32  hput; // next hbag push index
   u32  rput; // next rbag push index
   u32  sidx; // steal index
-  u32  nloc[32]; // node allocation indices
-  u32  vloc[32]; // vars allocation indices
+  u32  nloc[0xFFF]; // node allocation indices
+  u32  vloc[0xFFF]; // vars allocation indices
   Pair hbag_buf[HLEN]; // high-priority redexes
 } TM;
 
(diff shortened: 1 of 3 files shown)


src/hvm.c
```diff
@@ -302,9 +307,18 @@ u32 global_sum(u32 x) {
...
 static inline u64 time64() {
+
+// if not on windows
+#ifndef _WIN32
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
   return (u64)ts.tv_sec * 1000000000ULL + (u64)ts.tv_nsec;
+#else
+  int64_t wintime;
+  GetSystemTimeAsFileTime((FILETIME*)&wintime);
```
developedby, line 318: We should use `QueryPerformanceCounter` here. GetSystemTimeAsFileTime is non-monotonic

developedby (changes requested):

src/hvm.c
```diff
@@ -113,9 +116,9 @@ typedef u32 Numb; // Numb ::= 29-bit (rounded up to u32)
 #define G_RBAG_LEN (TPC * RLEN)
 
 typedef struct Net {
-  APair node_buf[G_NODE_LEN]; // global node buffer
-  APort vars_buf[G_VARS_LEN]; // global vars buffer
-  APair rbag_buf[G_RBAG_LEN]; // global rbag buffer
+  APair* node_buf; // global node buffer, size = G_NODE_LEN
```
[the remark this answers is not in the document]
sepcnt: Since large array will cause '[Total size of array must not exceed 0x7fffffff bytes](https://stackoverflow.com/questions/42173040/error-total-size-of-array-must-not-exceed-0x7fffffff-bytes)' on Windows.

src/hvm.c
sepcnt: If targeting at at least Windows 8, GetSystemTimeAsFileTime is much better I think.

src/hvm.c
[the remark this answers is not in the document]
developedby: But that's for static arrays, here Net and GNet are dinamically allocated so this shouldn't apply. 
Was this giving you a compilation error?

src/hvm.c
```diff
@@ -300,11 +305,37 @@ u32 global_sum(u32 x) {
   return sum;
 }
 
+#ifdef _WIN32
+static int64_t _win_time_offset = 0;
+static LARGE_INTEGER _win_time_freq, _win_perf_offset;
+
+static inline void _win_start_timer() {
+  QueryPerformanceFrequency(&_win_time_freq);
```
developedby, line 313: Why not just use the performance counter directly? We dont care about system time, this is just a timer.
```
#ifdef _WIN32
static inline u64 time64() {
  LARGE_INTEGER freq;
  LARGE_INTEGER counter;
  QueryPerformanceFrequency(&freq);
  QueryPerformanceCounter(&counter);
  return (u64)((counter.QuadPart * 1000000000) / freq.QuadPart);
}
#else
...
```
This will return a monotonic nanosecond timestamp like the unix version does
(i haven't tested this, but the internet tells me this should work)

PR #304, opened 2024-05-19
Create cmake-multi-platform.yml

SYNAX07:


developedby:
What is this file used for? We don't use cmake for hvm

PR #305, opened 2024-05-19
Fix long function names

### LINKED ISSUE #303: Function names longer than 32 bytes don't work
https://github.com/HigherOrderCO/HVM/blob/542ca9884428924c559beeeda2786abaec81defa/src/hvm.rs#L930 underflows for function names longer than 32 bytes, causing an immediate crash if HVM is compiled in debug mode, and inscrutable errors later if it's compiled in release mode.

CatsAreFluffy:
Fixes #303. The threshold for truncating names is 31, not 32, since you need a byte for a null terminator.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -113,7 +113,7 @@ checksum = "809e18805660d7b6b2e2b9f316a5099521b5998d5cba4dda11b5157a21aaef03"
 
 [[package]]
 name = "hvm"
-version = "2.0.9"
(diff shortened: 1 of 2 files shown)


VictorTaelin:
We should increase the name size to 256 or so though

enricozb:
Closing in favor of https://github.com/HigherOrderCO/HVM/pull/317, which increases function name limit to 256.

PR #313, opened 2024-05-20
Expose modules as a library

eduhenke:
This PR exposes the inner modules so that other projects can use HVM as a library. More focus on designing stabler APIs would be interesting in the future.


developedby (approved):
Looks good, this will also be useful for Bend

kings177 (approved):

PR #317, opened 2024-05-21
Fix CUDA compilation, add SHL/SHR

### LINKED ISSUE #285: HVM panics when the input has the number 0.0
As reported in https://github.com/HigherOrderCO/Bend/issues/334 .
Will be fixed once we get the new implementation of F24

VictorTaelin:
Fixes https://github.com/HigherOrderCO/HVM/issues/285, https://github.com/HigherOrderCO/HVM/issues/303


src/ast.rs
```diff
@@ -231,62 +223,74 @@ impl Numb {
...
-        if numb.get_flp() {
-          format!(":{:.3}", val)
+        if val.is_infinite() {
+          if val.is_sign_positive() {
+            format!("+inf")
+          } else {
+            format!("-inf")
+          }
+        } else if val.is_nan() {
+          format!("+NaN")
```
enricozb, line 265: Why is this `+NaN`?

src/hvm.rs
```diff
@@ -216,163 +220,151 @@ impl Numb {
...
+          OP_REM => Numb::new_i24(av.wrapping_rem(bv)),
+          FP_REM => Numb::new_i24(bv.wrapping_rem(av)),
+          OP_EQ  => Numb::new_i24((av == bv) as i32),
+          OP_NEQ => Numb::new_i24((av != bv) as i32),
+          OP_LT  => Numb::new_i24((av <  bv) as i32),
+          OP_GT  => Numb::new_i24((av >  bv) as i32),
+          OP_AND => Numb::new_i24(av & bv),
+          OP_OR  => Numb::new_i24(av | bv),
+          OP_XOR => Numb::new_i24(av ^ bv),
+          _      => unreachable!(),
```
enricozb, line 345: Do we want to panic here for doing shifts on `i24`? Same question for `f24`.

enricozb (approved):

src/ast.rs
```diff
@@ -41,98 +40,83 @@ new_parser!(CoreParser);
...
+      _ if self.try_consume("%")   => hvm::OP_REM,
+      _ if self.try_consume(":%")  => hvm::FP_REM,
+      _ if self.try_consume("=")   => hvm::OP_EQ,
+      _ if self.try_consume("!")   => hvm::OP_NEQ,
+      _ if self.try_consume("<")   => hvm::OP_LT,
+      _ if self.try_consume(">")   => hvm::OP_GT,
+      _ if self.try_consume("&")   => hvm::OP_AND,
+      _ if self.try_consume("|")   => hvm::OP_OR,
+      _ if self.try_consume("^")   => hvm::OP_XOR,
+      _ if self.try_consume("<<")  => hvm::OP_SHL,
```
developedby, line 63: This has to come before parsing "<"

src/ast.rs
```diff
@@ -41,98 +40,83 @@ new_parser!(CoreParser);
...
+      _ if self.try_consume(":<<") => hvm::FP_SHL,
+      _ if self.try_consume(">>")  => hvm::OP_SHR,
```
developedby, line 65: This has to come before parsing ">"

nicolas-abril (changes requested):

src/ast.rs
enricozb: Good catch

src/hvm.rs
VictorTaelin: will think about it!

PR #326, opened 2024-05-21
cargo clippy fix

Liberxue:
# PR Description

## Changes

Fixed issues flagged by Clippy for unnecessary type casting, such as casting from u32 to u32.

Replaced unnecessary usage of format! with direct string concatenation.

Removed redundant return statements to simplify function returns.

Replaced manual variable swapping with std::mem::swap to enhance code readability and conciseness.

## Details

Updated instances where single-character strings were pushed to a string, using push(char) instead of push_str.

Improved performance and readability by using std::mem::swap for swapping variables.

Ensured compliance with Rust's Clippy linter for better code quality and maintenance.


enricozb:
thanks but in this repo specifically we are not using a code formatter as we're supporting three separate runtimes, and want the code to look the same across all there languages (c, cuda, rust).

PR #327, opened 2024-05-21
add test suite

tjjfvi:
- added a test suite that automatically tests every HVM file in `examples/` and `tests/programs/` across all enabled runtimes
- added a CI workflow to build & test every push to a PR
- adjusted all of the `examples/` programs to run in a reasonable amount of time by default
- fixed several bugs uncovered in testing:
  - fixed C/CUDA implementation of I24
  - fixed a bug with `new_f24` that "rounded" some NaN patterns to 0
  - fixed C/CUDA readback of large numbers like 16777215 (which were being printed as random characters instead)


examples/sum_rec/main.hvm
```diff
@@ -1,8 +1,8 @@
 @main = a
-  & @sum ~ (28 (0 a))
+  & @sum ~ (24 (0 a))
 
 @sum = (?(((a a) @sum__C0) b) b)
 
 @sum__C0 = ({c a} ({$([*2] $([+1] d)) $([*2] $([+0] b))} f))
-  &! @sum ~ (a (b $(:[+] $(e f))))
```
enricozb, line 7: Not sure if we should fix these manually here or coordinate with the bend team to use the new syntax.

tjjfvi: probably both; although I think the latter may already have been done?

VictorTaelin:
Thanks, I can't review this right now but will ASAP.

VictorTaelin:
Seems like this doesn't test the compiled versions yet, right? It should be good to add that later, as the compiled functions are a major source of bugs.

VictorTaelin (approved):

PR #336, opened 2024-05-22
IO/FFI Standard

VictorTaelin:
This adds a much simpler IO functionality, based on external FFI calls. The commit message describes how it works.

@developedby This also has a new `demo_io` file which should replace the current built-in IO on Bend. I'm calling it `IO_T` to avoid conflict, but it should be just `IO`. I've designed it in a way that won't need to be changed as we add more IOs (and should be extensive later). There is also the issue of Bend consuming the output. Perhaps that's a moment for us to rework the output into something cleaner?

@enricozb Could you kindly review and implement on CUDA? The old commit has most of the core functions needed (like reading needed data from GPU memory to CPU). So you can use these to implement the new standard.


developedby:
What I've thought of doing to print IO but not the hvm result output is having HVM end with a very unique magic keyword.

On Bend's side I can buffer stdout a little bit to check for a new line followed by that magic keyword and capture only from that line onwards (to also capture any additional information like runtime stats).

It could be for example "\nHVM Result:", but it has to be something a user would never accidentally type.

PR #337, opened 2024-05-22
Fix i24 comparison

VictorTaelin:


enricozb (changes requested):
This bug is still present in both the C and CUDA impls.

enricozb:
Also needs to do the `cargo insta review` to update the snapshots of the tests (as long as they look sane obv)

enricozb (approved):

PR #341, opened 2024-05-23
HVM cli arg version

miolini:
Hi!

This pull request introduces a new command-line argument -V to the HVM Rust project. The purpose of this argument is to print the current version of the application using the value specified in the Cargo package.

Changes:
    Added a new CLI argument -V.
    Implemented functionality to fetch and display the version from the Cargo package.

Details:
    Modified src/main.rs to include the -V argument.
    Utilized the env! macro to retrieve the version from Cargo.toml.
    Updated the help message to include the new -V option.

Usage:
    Running hvm -V will now print the current version of the application.

Motivation:
Adding the -V argument enhances the user experience by providing an easy way to check the application version directly from the command line. This aligns with common CLI conventions and improves usability.

Notes:
    No breaking changes introduced.

Please review the changes and provide feedback. 

Thank you!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,6 +22,7 @@ extern "C" {
 fn main() {
   let matches = Command::new("hvm")
     .about("HVM2: Higher-order Virtual Machine 2 (32-bit Version)")
+    .version(env!("CARGO_PKG_VERSION"))
     .subcommand_required(true)
     .arg_required_else_help(true)
     .subcommand(


NoamDev:
Disclaimer: Not a team member.
Great Idea! This would be great to clarify bug reports as well.
I think `--version` would be better, this can be done by using `long_version()` instead.
This what is used by gcc, clang, nvcc, node.
python actually supports both, which is always a possibility.
However, might be better to reserve `-v` for verbose mode which might be added one day. having both `-V` and `-v` might be confusing.

developedby:
> Disclaimer: Not a team member. Great Idea! This would be great to clarify bug reports as well. I think `--version` would be better, this can be done by using `long_version()` instead. This what is used by gcc, clang, nvcc, node. python actually supports both, which is always a possibility. However, might be better to reserve `-v` for verbose mode which might be added one day. having both `-V` and `-v` might be confusing.

The `version` argument adds both short and long options.

developedby (approved):

kings177 (approved):
thank you!

miolini:
Thank you!

PR #346, opened 2024-05-24
error gracefully when `@main` is missing

### LINKED ISSUE #344: programs without `@main` crash Rust runtime and hang C runtime
This is what's currently causing tests to timeout on main; `tuples.hvm` is empty.

```
$ hvm run empty.hvm
thread 'main' panicked at src/main.rs:177:69:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
```
$ hvm run-c empty.hvm
[hangs]
```

tjjfvi:
Fixes #344

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -505,18 +505,19 @@ impl Book {
     }
     let mut book = hvm::Book { defs: Vec::new() };
     for (fid, name) in &fid_to_name {
-      if let Some(ast_def) = self.defs.get(name) {
-        let mut def = hvm::Def {
-          name: name.clone(),
-          safe: true,
-          root: hvm::Port(0),
-          rbag: vec![],
-          node: vec![],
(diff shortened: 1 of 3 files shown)


src/ast.rs
```diff
@@ -505,18 +505,19 @@ impl Book {
...
-          rbag: vec![],
-          node: vec![],
-          vars: 0,
-        };
-        ast_def.build(&mut def, &name_to_fid, &mut BTreeMap::new());
-        book.defs.push(def);
-      }
+      let Some(ast_def) = self.defs.get(name) else {
+        panic!("missing `@main` definition");
+      };
```
enricozb, line 510: maybe just `expect` instead of `let .. else`?

enricozb (approved):

PR #347, opened 2024-05-24
CUDA Basic IO

enricozb:
Adds support for the `PUT_TEXT`, `GET_TIME`, and `PUT_TIME` IO operations.


src/hvm.cu
```diff
@@ -195,10 +195,20 @@ struct Def {
...
+  char name[256];
+  Port (*func)(GNet*, Port);
+} FFn;
+
 // Book of Definitions
 struct Book {
   u32 defs_len;
   Def defs_buf[0x4000]; // 256 MB
+  u32 ffns_len;
+  FFn ffns_buf[0x4000]; // 256 MB
```
VictorTaelin, line 211: these sizes look right. defs are larger than fns. why same size?

src/hvm.cu
```diff
@@ -1802,16 +1827,229 @@ Port gnet_expand(GNet* gnet, Port port) {
...
+    if (get_tag(app_port) != CON) break;
+    app_node = gnet_node_load(gnet, get_val(app_port));
+    arg_port = gnet_expand(gnet, get_fst(app_node));
+    ctr.args_buf[ctr.args_len++] = arg_port;
+  }
+
+  return ctr;
+}
+
+// Reads back a UTF-16 string.
```
VictorTaelin, line 1874: strings in hvm are actually just lists of unicode codepoints, we must update that

VictorTaelin (approved):

src/hvm.cu
enricozb: Sorry, good catch

PR #354, opened 2024-05-27
Updates checks.yml to work after rust-toolchain was removed

kings177:


.github/workflows/checks.yml
```diff
@@ -13,7 +13,9 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v3
-      - uses: dsherret/rust-toolchain-file@v1
+        with:
+          profile: minimal
+          toolchain: 1.74.0
```
tjjfvi, line 18: These blocks are being passed to `checkout` and aren't doing anything.

The ubuntu image comes with rust 1.78 installed [by default](https://github.com/actions/runner-images/blob/ubuntu22/20240516.1/images/ubuntu/Ubuntu2204-Readme.md#rust-tools), so these can be removed

kings177: oh, true, i just brain deadly added them without much thought. i put the 1.74 just because it is the same as in Bend's minimum working version.

tjjfvi (approved):

Issue #324, opened 2024-05-21
Option to limit the number of threads ?


khatharsis42:
No matter how you run things, it will always run the maximum number of threads on your machine. This is not always the desired behaviour. It'd be nice if it could be changed, with perhaps an equivalent to OMP_NUM_THREADS for OpenMP ?

kings177:
For now, we don't have the functionality of changing the number of max threads yet, however, you can change it in the code inside `build.rs`, changing `let tpcl2 = (cores as f64).log2().floor() as u32;` to let's say `let tpcl2 = 3` for example, would run with 8 threads max (threads = 2^tpcl2)

khatharsis42:
BTW, it might be a good idea to make it so it doesn't use all the *threads* the system has, but rather all the *cores*. Especially since using too many threads seems to lead to some OOM errors ?

Janiczek:
What's the reason the number of cores is always a power of two?

khatharsis42:
@Janiczek a lot of developper have a superstitious belief the having stuff be base two makes the code go faster. In the case of threads, it doesn't make much sense, you should be more concerned about thread placement (ie not putting multiples threads running on the same core) and not overpopulating (ie using more threads than you have *cores*).
This can be tricky, because most cores nowadays have hyperthreading enabled, meaning they can run multiple threads at once, albeit at decreased efficiency.

VictorTaelin:
HigherOrderCO's repositories are being reorganized: HVM1 and HVM2 swapped repository names, so that each holds its own code and history. We are closing every open issue and pull request as part of the move: if this one is still relevant, please re-open it on the matching repository, https://github.com/HigherOrderCO/HVM1 or https://github.com/HigherOrderCO/HVM2.

PR #358, opened 2024-05-28
Dynamic number of threads using HVM_NUM_THREADS.

khatharsis42:
I changed a bit of src/hvm.c to allow the user to dynamically change the number of threads when running. This is especially useful right now on computers that have a lot of threads, since you might not want to use all of them. Related to https://github.com/HigherOrderCO/HVM/issues/324


VictorTaelin:

PR #359, opened 2024-05-28
use `dyntest` for tests

tjjfvi:
This lets us create a distinct test case for each file in `examples`/`test/programs`, which speeds up the tests from 5s to 2s, makes the test runner output much more informative, and gives more powerful command-line filtering.


tjjfvi:
Ah, forgot that it needs nightly.

PR #361, opened 2024-05-29
Numeric Casting Operations

### LINKED ISSUE #322: Numeric Type Conversion
**Is your feature request related to a problem? Please describe.**
Currently to convert from integer to float we have to do an expensive bit-by-bit transformation, over 30 interactions.
It'd be nice if the HVM either exposed the native casting operation or provided operations that allowed for this to be done cheaply.

**Describe the solution you'd like**
We could define the interaction `SYM(TYP1) ~ OP(NUM(TYP2) x)` to produce `NUM(TYP1) ~ x`, doing the appropriate casting.
Currently it produces `0 ~ x`.
For float to integer it should truncate the value.

**Describe alternatives you've considered**
An alternative that doesn't require changing the OPER interaction semantics, would be to add a CLZ (count leading zeros) operation and a rounding/truncating operation.

CLZ allows converting uint to float cheaply by directly providing a way to get the exponent.
Int to float requires flipping the bits first, but it's basically the same.
However, this requires duplicating the number to convert to a float (one copy for exponent, one for mantissa), so not ideal.

A rounding operation converts from float to int and then we can easily convert to uint.

**Additional context**
There are many algorithms that work by incrementing a counter and then using that counter in a float operations. This could be done by incrementing a float, but it stops working at some point.
It also makes comparing the integer values a problem, since they're now floats and have intrinsic errors.

enricozb:
Adds support for numeric casting operations, `[u24]`, `[i24]`, `[f24]`.

Fixes https://github.com/HigherOrderCO/HVM/issues/322.

Casts are always defined, and spiritually follow rusts numeric casting semantics:
- `i24 <-> u24` is just reinterpretation of bits.
- `f24 -> i24` or `f24 -> u24` casts to the "closest" integer representing this float, saturating if out of range and `0` if `NaN`.
- `i24 -> f24` or `u24 -> f24` casts to the "closest" float representing this integer.


src/ast.rs
```diff
@@ -264,7 +283,7 @@ impl Numb {
         } else if val.is_nan() {
           format!("+NaN")
         } else {
-          format!("{:?}", val)
```
tjjfvi, line 267: why this change?

src/hvm.rs
```diff
@@ -266,16 +266,47 @@ impl Numb {
...
+  //   u24  -> f24: casts to the "closest" float representing this integer.
+  pub fn cast(a: Self, b: Self) -> Self {
+    match (a.get_sym(), b.get_typ()) {
+      (TY_U24, TY_U24) => b,
+      (TY_U24, TY_I24) => Self::new_u24(b.get_i24() as u32),
+      (TY_U24, TY_F24) => Self::new_u24(b.get_f24().clamp(0.0, 16777215.0) as u32),
+
+      (TY_I24, TY_U24) => Self::new_i24(b.get_u24() as i32),
+      (TY_I24, TY_I24) => b,
+      (TY_I24, TY_F24) => Self::new_i24(b.get_f24().clamp(-8388608.0, 8388607.0) as i32),
```
tjjfvi, line 302: might it be somewhat more efficient to clamp it as an integer? not that it matters much

src/hvm.rs
```diff
@@ -266,16 +266,47 @@ impl Numb {
...
+  // - f24  -> i24,
+  //   f24  -> u24: casts to the "closest" integer representing this float,
+  //                saturating if out of range and 0 if NaN
+  // - i24  -> f24,
+  //   u24  -> f24: casts to the "closest" float representing this integer.
+  pub fn cast(a: Self, b: Self) -> Self {
+    match (a.get_sym(), b.get_typ()) {
+      (TY_U24, TY_U24) => b,
+      (TY_U24, TY_I24) => Self::new_u24(b.get_i24() as u32),
+      (TY_U24, TY_F24) => Self::new_u24(b.get_f24().clamp(0.0, 16777215.0) as u32),
```
tjjfvi, line 298: maybe extract these numbers to constants? if they were clamped as integers you could have `const U24_MAX = 1 << 24 - 1;`

tests/snapshots/run__file@empty.hvm.snap
```diff
@@ -4,6 +4,6 @@ expression: rust_output
 input_file: tests/programs/empty.hvm
 ---
 exit status: 101
-thread 'main' panicked at src/ast.rs:508:41:
+thread 'main' panicked at src/ast.rs:527:41:
```
tjjfvi, line 7: off-topic but this diff is pretty annoying; we should figure out a better way to test this

src/ast.rs
enricozb: whoops, was testing to see if there were format specifiers where Rust and C both output the same float representation

src/hvm.rs
enricozb: both good ideas. I clamped on floats in the rust impl b/c that's what we have to do in the C impls, as casting float -> int out of the range of the int type is UB. But as a reference implementation I think it's it might be clearer with the integer cast first.

src/hvm.rs
enricozb: Did this for the rust impl, but the other runtimes can't have it clamped as an integer unfortunately

tjjfvi (approved):

PR #363, opened 2024-05-30
add open, close, read, write

enricozb:
Adds `OPEN_FILE`, `CLOSE_FILE`, `READ_LINE`, `READ_CHAR`, and `WRITE` IO functions. Since the read and write operations take in a file descriptor, I've also reserved numbers `0` `1` and `2` to have their usual FD meaning, so users can print to stderr, read from stdin, etc.

Here's an example:
```python
type IO_T:
  Done { magic, expr }
  Call { magic, func, argm, cont }

def IO_T/MAGIC:
  return (0xD0CA11, 0xFF1FF1)

def IO_T/wrap(x):
  return IO_T/Done(IO_T/MAGIC, x)

def IO_T/bind(a, b):
  match a:
    case IO_T/Done:
      return b(a.expr)
    case IO_T/Call:
      return IO_T/Call(IO_T/MAGIC, a.func, a.argm, lambda x: IO_T/bind(a.cont(x), b))

def call_io(func, argm):
  return IO_T/Call(IO_T/MAGIC, func, argm, lambda x: IO_T/Done(IO_T/MAGIC, x))

def main:
  do IO_T:
    * <- call_io("WRITE", (1, "what is your name?\n"))
    name <- call_io("READ_LINE", 0)
    * <- call_io("WRITE", (1, "Hello, '"))
    * <- call_io("WRITE", (1, name))
    * <- call_io("WRITE", (1, "'!\n"))

    file <- call_io("OPEN_FILE", ("testing.txt", "w"))
    * <- call_io("WRITE", (file, "Your name is: "))
    * <- call_io("WRITE", (file, name))
    * <- call_io("WRITE", (file, "!"))
    * <- call_io("CLOSE_FILE", file)

    return 42
```


src/hvm.c
```diff
@@ -1240,7 +1235,7 @@ Port expand(Net* net, Book* book, Port port) {
 
 // Reads back a λ-Encoded constructor from device to host.
 // Encoding: λt ((((t TAG) arg0) arg1) ...)
-Ctr read_ctr(Net* net, Book* book, Port port) {
+Ctr port_to_ctr(Net* net, Book* book, Port port) {
```
VictorTaelin, line 1238: let's keep the readback functions prefixed by `read_` as a convention

VictorTaelin (approved):

src/hvm.c
enricozb: What about functions that are inverses of these? I named it `port_to_ctr` b/c I also created a `ctr_to_port`. What do you think we should call the functions that convert `Ctr` to `Port`?

VictorTaelin: I prefer `readback_x` and `inject_x` or `create_x` for the inverses, because it is consistent with how it is in other codes of the past.

VictorTaelin:
Is that example right? Isn't it `with` instead of `do`?

edusporto:
@VictorTaelin That example was fixed in #369, it should be correct now

VictorTaelin:
oh sure, my bad

PR #366, opened 2024-05-31
Type-conversion SYMs

CatsAreFluffy:
Add SYMs for the numeric types and SYM itself to the parser and the printers for each runtime, to allow fast conversions between numeric types. Note that this functionality was already present in the runtimes, it just wasn't previously accessible.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -45,6 +45,10 @@ impl<'i> CoreParser<'i> {
 
     // Parses the symbol
     let op = hvm::Numb::new_sym(match () {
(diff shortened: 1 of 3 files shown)


CatsAreFluffy:
Apparently #361 exists. Note that this PR's conversions between F24 and integer types are bitcasts rather than the numeric conversions #361 uses.

enricozb:
Yeah we're considering adding bit cast operators like `[as_f24]` if you just want to reinterpret the bits.

CatsAreFluffy:
#361 got merged and that conflicts with this so closing.

Issue #360, opened 2024-05-29
Segfault after printing results


nicolas-abril:
### Reproducing the behavior

Running this program with `hvm run-c` causes a segmentation fault after printing the results
```
@List/Cons = (a (b ((@List/Cons/tag (a (b c))) c)))

@List/Cons/tag = 1

@List/Nil = ((@List/Nil/tag a) a)

@List/Nil/tag = 0

@List/append = ((@List/append__C3 a) a)

@List/append__C0 = (* (c (d (a (b (* g))))))
  & @List/Cons ~ (a (f g))
  & @List/append ~ (b (e f))
  & @List/Cons ~ (c (d e))

@List/append__C1 = (?(((* (* (a a))) @List/append__C0) b) b)

@List/append__C2 = (* ({a c} ({b d} ((@List/append__C1 (a (b (e f)))) f))))
  & @List/Cons ~ (c (d e))

@List/append__C3 = (?(((a a) @List/append__C2) b) b)

@gen_lst = a
  & @gen_lst__bend0 ~ a

@gen_lst__bend0 = (a ({$([<0] ?(((* (* @List/Nil)) @gen_lst__bend0__C0) (a (b c)))) b} c))

@gen_lst__bend0__C0 = (* ($([%] $(a $([*7] $([+3] {b $([:%5] $([+3] c))})))) ({$([:-1] d) a} f)))
  & @List/Cons ~ (b (e f))
  & @gen_lst__bend0 ~ (c (d e))

@main = (* a)
  & @main2 ~ (32 a)

@main2 = (a (b d))
  & @gen_lst ~ (31 (a {b c}))
  & @quickSort ~ (c d)

@partition = a
  & @partition__fold0 ~ a

@partition__fold0 = ((@partition__fold0__C3 a) a)

@partition__fold0__C0 = (b (a (c (a d))))
  & @List/Cons ~ (b (c d))

@partition__fold0__C1 = (* (a (b (d (c d)))))
  & @List/Cons ~ (a (b c))

@partition__fold0__C2 = (* ({c d} (a ({b $([>] $(c ?((@partition__fold0__C0 @partition__fold0__C1) (d (e (f g))))))} g))))
  & @partition__fold0 ~ (a (b (e f)))

@partition__fold0__C3 = (?(((* (@List/Nil @List/Nil)) @partition__fold0__C2) a) a)

@quickSort = ((@quickSort__C1 a) a)

@quickSort__C0 = (* ({b e} (a i)))
[...]

PR #368, opened 2024-06-03
recursive pretty print

enricozb:
Fixes issues with segfault after printing: https://github.com/HigherOrderCO/HVM/issues/360.
Also it doesn't overload the `ERA` node.

Of course a large enough net will still overflow the stack. On my machine, this example bend code:
```python
def create(n):
  bend x = 0:
    when x < n:
      str = List/Cons { head: 1, tail: fork(x + 1) }
    else:
      str = List/Nil
  return str

def main():
  return create(37)
```
Which previously segfaulted at `create(37)`, now segfaults somewhere between `create(50000)` and `create(100000)`.


VictorTaelin:
You just removed the stack? That's a regression, in CUDA this will cause the compiler to be unable to estimate the stack size of the pretty print function, which can have harsh consequences on performance elsewhere and cause segfaults in unrelated code. Have you tested it on the GPU?

enricozb:
I didn't test pretty print within any of the cuda code, didn't think of that as a use case. We can have a stacked version for debugging and use this for the output on the host then?

Issue #362, opened 2024-05-29
Eraser nodes being placed where they shouldn't


edusporto:
### Reproducing the behavior

As Nicolas pointed out a few days ago, some programs that should be equivalent return different results, with a few eraser nodes being placed in places they shouldn't. Nicolas suspects it's an issue with duplication, but I still haven't found it.

See the following program in Bend:
```python
map f xs = match xs {
  List/Cons: (List/Cons (f xs.head) (map f xs.tail))
  List/Nil: List/Nil
}

list = [1,2,3]
id = @x x

main = (map @f (map f list) [id])
```

It gets compiled to this, which works correctly:
```
@List/Cons = (a (b ((@List/Cons/tag (a (b c))) c)))

@List/Cons/tag = 1

@List/Nil = ((@List/Nil/tag a) a)

@List/Nil/tag = 0

@id = (a a)

@list = c
  & @List/Cons ~ (1 (b c))
  & @List/Cons ~ (2 (a b))
  & @List/Cons ~ (3 (@List/Nil a))

@main = b
  & @map ~ (@main__C0 (a b))
  & @List/Cons ~ (@id (@List/Nil a))

@main__C0 = (a b)
  & @map ~ (a (@list b))

@map = (a ((@map__C1 (a b)) b))

@map__C0 = (* (a (d ({(a b) c} f))))
  & @List/Cons ~ (b (e f))
  & @map ~ (c (d e))

@map__C1 = (?(((* @List/Nil) @map__C0) a) a)
```

Notice how `main` contains a call to `main__C0`. If we replace it with its value, we get
```
@List/Cons = (a (b ((@List/Cons/tag (a (b c))) c)))

@List/Cons/tag = 1

@List/Nil = ((@List/Nil/tag a) a)

@List/Nil/tag = 0

@id = (a a)

@list = c
  & @List/Cons ~ (1 (b c))
  & @List/Cons ~ (2 (a b))
  & @List/Cons ~ (3 (@List/Nil a))

@main = e
  & @map ~ ((a b) (d e))
  & @map ~ (a (@list b))
  & @List/Cons ~ (@id (@List/Nil d))

// @main__C0 = (a b)
//   & @map ~ (a (@list b))

@map = (a ((@map__C1 (a b)) b))

@map__C0 = (* (a (d ({(a b) c} f))))
  & @List/Cons ~ (b (e f))
  & @map ~ (c (d e))

[...]

PR #379, opened 2024-06-07
Propagate unsafety to definitions that depend on unsafe definitions

edusporto:
Related to #362.

This PR kind of fixes the issue, but not really. While this new version does return an error in the previously thought "correct" version of the program, which is expected since it actually performed unsafe operations, it doesn't error in the second version, which does the same. Returning an error in this version is much harder, since it's not cloning an unsafe definition, but an unsafe expression. According to Nicolas, that could be solved with an affine type system.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -2,7 +2,7 @@ use TSPL::{new_parser, Parser};
 use highlight_error::highlight_error;
 use crate::hvm;
 use std::fmt::{Debug, Display};
-use std::collections::{btree_map::Entry, BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet};
 
 // Types
 // -----
@@ -491,16 +491,22 @@ impl Tree {
   }
 
   pub fn direct_dependencies<'name>(&'name self) -> BTreeSet<&'name str> {
-    match self {
-      Tree::Ref { nam } => BTreeSet::from([nam.as_str()]),
-      Tree::Con { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
-      Tree::Dup { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
-      Tree::Opr { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
-      Tree::Swi { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
-      Tree::Num { val } => BTreeSet::new(),
-      Tree::Var { nam } => BTreeSet::new(),
-      Tree::Era => BTreeSet::new(),
+    let mut stack: Vec<&Tree> = vec![self];
+    let mut acc: BTreeSet<&'name str> = BTreeSet::new();
+    
+    while let Some(curr) = stack.pop() {
+      match curr {
+        Tree::Ref { nam } => { acc.insert(nam); },
+        Tree::Con { fst, snd } => { stack.push(fst); stack.push(snd); },
+        Tree::Dup { fst, snd } => { stack.push(fst); stack.push(snd); },
+        Tree::Opr { fst, snd } => { stack.push(fst); stack.push(snd); },
+        Tree::Swi { fst, snd } => { stack.push(fst); stack.push(snd); },
+        Tree::Num { val } => {},
+        Tree::Var { nam } => {},
+        Tree::Era => {},
+      };
     }
+    acc
   }
 }
 
@@ -537,7 +543,6 @@ impl Book {
       }
     }
     let mut book = hvm::Book { defs: Vec::new() };
-    let mut lookup = BTreeMap::new();
     for (fid, name) in &fid_to_name {
       let ast_def = self.defs.get(name).expect("missing `@main` definition");
       let mut def = hvm::Def {
@@ -550,9 +555,8 @@ impl Book {
       };
       ast_def.build(&mut def, &name_to_fid, &mut BTreeMap::new());
       book.defs.push(def);
-      lookup.insert(name.clone(), book.defs.len() - 1);
     }
-    self.propagate_safety(&mut book, &lookup);
+    self.propagate_safety(&mut book, &name_to_fid);
     return book;
   }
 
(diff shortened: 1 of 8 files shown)


src/ast.rs
```diff
@@ -488,6 +489,19 @@ impl Tree {
...
+      Tree::Ref { nam } => BTreeSet::from([nam.as_str()]),
+      Tree::Con { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
+      Tree::Dup { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
+      Tree::Opr { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
+      Tree::Swi { fst, snd } => &fst.direct_dependencies() | &snd.direct_dependencies(),
+      Tree::Num { val } => BTreeSet::new(),
+      Tree::Var { nam } => BTreeSet::new(),
+      Tree::Era => BTreeSet::new(),
+    }
+  }
```
edusporto, line 504: Writing a recursive function like this may not be the best idea since we still don't have a way to increase stack size as needed like in the `hvm-64` repo. I think it's fine for now to add this function like this, since pretty much all the other functions related to trees are also recursive and if this one fails, they do too.

tests/run.rs
```diff
@@ -45,6 +45,11 @@ fn test_file(path: &Path) {
   let rust_output = execute_hvm(&["run".as_ref(), path.as_os_str()], false).unwrap();
   assert_snapshot!(rust_output);
 
+  if contents.contains("@test-rust-only = 1") {
+    println!("only testing rust implementation for {path:?}");
+    return;
+  }
+
```
edusporto, line 52: I added the possibility to only run tests in the Rust implementation, since the C one doesn't throw an error in this same case. This is something to discuss - should the C and CUDA implementations also "panic" printing this error and stopping the program?

src/ast.rs
enricozb: if it becomes a problem we can just write this imperatively

tests/run.rs
enricozb: I think this is fine as is, and is in line with the other `test-io` pattern.

enricozb (commented):
So if I understand correctly, this doesn't account for user-introduced unsafety? It only accounts for when users reference an unsafe function?

edusporto:
> So if I understand correctly, this doesn't account for user-introduced unsafety? It only accounts for when users reference an unsafe function?

Depends on what you mean by "user-introduced", since users also introduce definitions. It will only catch unsafe operations if the unsafe operation is contained in a definition, and a reference to that definition is directly duplicated.

For example, exponentiation of church naturals (which I think should behave poorly in IC?)
```python
unsf = @f @x (f (f x))
main = (unsf unsf)
```

This currently gets compiled to
```
@main = a
  & @unsf ~ (@unsf a)

@unsf = ({(b c) (a b)} (a c))
```

which is correctly marked as unsafe. We could also do
```
@main = a
  & ({(b c) (d b)} (d c)) ~ (({(b c) (d b)} (d c)) a)
```
which is not marked as unsafe, and does not terminate (loops forever)

edusporto:
@developedby this PR breaks (surprisingly only) one of the tests in Bend, `guide_shader_dummy.bend`:
```python
# given a shader, returns a square image
def render(depth, shader):
  bend d = 0, i = 0:
    when d < depth:
      color = (fork(d+1, i*2+0), fork(d+1, i*2+1))
    else:
      width = depth / 2
      color = shader(i % width, i / width)
  return color

# given a position, returns a color
# for this demo, it just busy loops
def demo_shader(x, y):
  bend i = 0:
    when i < 10:
      color = fork(i + 1)
    else:
      color = 0x000001
  return color

# renders a 256x256 image using demo_shader
def main:
  return render(5, demo_shader)
```
`demo_shader` is marked as unsafe since it duplicates `i`, and the program fails with the "ERROR: attempt to clone a non-affine global reference." error when `render` duplicates that function.

This is a very natural program to write, and the safety check this PR introduces will probably break many more programs created by users of Bend. I'm not sure how to deal with this... Should we really add this check in the way it's done by this PR, or should we do something else?

developedby:
> @developedby this PR breaks (surprisingly only) one of the tests in Bend, `guide_shader_dummy.bend`:
> 
> ```python
> # given a shader, returns a square image
> def render(depth, shader):
>   bend d = 0, i = 0:
>     when d < depth:
>       color = (fork(d+1, i*2+0), fork(d+1, i*2+1))
>     else:
>       width = depth / 2
>       color = shader(i % width, i / width)
>   return color
> 
> # given a position, returns a color
> # for this demo, it just busy loops
> def demo_shader(x, y):
>   bend i = 0:
>     when i < 10:
>       color = fork(i + 1)
>     else:
>       color = 0x000001
>   return color
> 
> # renders a 256x256 image using demo_shader
> def main:
>   return render(5, demo_shader)
> ```
> 
> `demo_shader` is marked as unsafe since it duplicates `i`, and the program fails with the "ERROR: attempt to clone a non-affine global reference." error when `render` duplicates that function.
> 
> This is a very natural program to write, and the safety check this PR introduces will probably break many more programs created by users of Bend. I'm not sure how to deal with this... Should we really add this check in the way it's done by this PR, or should we do something else?

Yes, this is the intended behaviour and a big limitation of hvm32

developedby:
If we plan on continuing with hvm32 for a lot longer, I should probably add some monomorphization system to Bend to avoid hitting this limitation so often

src/ast.rs
tjjfvi: This function is quadratic, due to repeatedly cloning these btreesets; `(@a (@b (@c (@d (@e ...)))))` will take triangular time (which is asymptotically quadratic).

src/ast.rs
```diff
@@ -523,6 +537,7 @@ impl Book {
       }
     }
     let mut book = hvm::Book { defs: Vec::new() };
+    let mut lookup = BTreeMap::new();
```
tjjfvi, line 540: isn't this duplicative of `name_to_fid`?

src/ast.rs
```diff
@@ -535,7 +550,86 @@ impl Book {
...
+
+    let mut process = |tree: &'name Tree, name: &'name str| {
+      for dependency in tree.direct_dependencies() {
+        match result.entry(dependency) {
+          Entry::Vacant(_) => panic!("global definition depends on undeclared reference"),
+          Entry::Occupied(mut entry) => {
+            // dependency => name
+            entry.get_mut().insert(name);
+          },
+        }
```
tjjfvi, line 622:
```suggestion
        result
          .get_mut()
          .expect("global definition depends on undeclared reference")
          .insert(name);
```

src/ast.rs
```diff
@@ -535,7 +550,86 @@ impl Book {
...
+      if !def.safe {
+        stack.push(&name);
+      }
+    }
+
+    while let Some(curr) = stack.pop() {
+      if visited.contains(curr) {
+        continue;
+      }
+      visited.insert(curr);
```
tjjfvi, line 585:
```suggestion
      if !visited.insert(curr) {
        continue;
      }
```

tjjfvi: Though you could probably just use `def.safe` for tracking this

edusporto:
All suggestions were resolved
Any other details or should we merge it? @enricozb @developedby @tjjfvi

tjjfvi:
fine by me, although slight nit: `direct_dependencies_reversed` could instead be named `direct_dependents` (with similar naming changes made elsewhere)

enricozb (approved):

PR #386, opened 2024-06-10
add io_seek, operate on bytes instead of strings

enricozb:
- Adds a `SEEK` io operation.
- replace `READ_LINE` and `READ_CHAR` with just a `READ`, taking in the number of bytes to be read.
- `READ` and `WRITE` now operate on lists of bytes

### DIFF
--- a/src/run.c
+++ b/src/run.c
@@ -256,20 +256,20 @@ FILE* readback_file(Port port) {
 // Reads from a file a specified number of bytes.
 // `argm` is a tuple of (file_descriptor, num_bytes).
 Port io_read(Net* net, Book* book, Port argm) {
-  if (get_tag(peek(net, argm)) != CON) {
-    fprintf(stderr, "io_read: expected tuple, but got %u\n", get_tag(peek(net, argm)));
+  if (get_tag(argm) != CON) {
+    fprintf(stderr, "io_read: expected tuple, but got %u\n", get_tag(argm));
     return new_port(ERA, 0);
   }
 
   Pair args = node_load(net, get_val(argm));
+  Port file = expand(net, book, get_fst(args));
+  u32 num_bytes = get_u24(get_val(expand(net, book, get_snd(args))));
 
-  FILE* fp = readback_file(peek(net, get_fst(args)));
+  FILE* fp = readback_file(file);
   if (fp == NULL) {
     fprintf(stderr, "io_read: invalid file descriptor\n");
     return new_port(ERA, 0);
   }
-
-  u32 num_bytes = get_u24(get_val(peek(net, get_snd(args))));
 
   /// Read a string.
   Bytes bytes;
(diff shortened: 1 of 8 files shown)


src/run.c
```diff
@@ -213,55 +253,47 @@ FILE* readback_file(Port port) {
...
 }
 
-// Reads a single char from `argm`.
-Port io_read_char(Net* net, Book* book, Port argm) {
-  FILE* fp = readback_file(peek(net, argm));
-  if (fp == NULL) {
+// Reads from a file a specified number of bytes.
+// `argm` is a tuple of (file_descriptor, num_bytes).
+Port io_read(Net* net, Book* book, Port argm) {
+  if (get_tag(peek(net, argm)) != CON) {
```
developedby, line 259: `peek` is not enough here and in all the other places. If we pass a constant hidden behind a `REF` node these IO functions will fail.

For example, if in Bend I call open like this:
```
(IO/call IO/MAGIC "OPEN" (MyConstFile, "r") @x x)
```
Where `MyConstFile` is a reference to a function that contains the file path.

Instead of `peek` we should use a different function that has a loop of peek and expand. This has the problem of looping infinitely if the user accidentally passes a recursive ref.

developedby (changes requested):

src/run.c
enricozb: switched to using `expand` for every arg of every io func

src/run.c
```diff
@@ -276,52 +308,100 @@ Port io_open_file(Net* net, Book* book, Port argm) {
...
+//    - 0 (SEEK_SET): beginning of file
+//    - 1 (SEEK_CUR): current position of the file pointer
+//    - 2 (SEEK_END): end of the file
+Port io_seek(Net* net, Book* book, Port argm) {
+  if (get_tag(argm) != CON) {
+    fprintf(stderr, "io_seek: expected first tuple, but got %u\n", get_tag(argm));
+    return new_port(ERA, 0);
+  }
+
+  Pair args1 = node_load(net, get_val(argm));
```
developedby, line 374: Every time we read an argument node we want to call expand no?
So here `(IO/call "SEEK" SeekArgsRef)` where SeekArgsRef is not a variable or a tuple, but a reference to a definition, we'll still get the error no?

In `readback_ctr` this is already being done for example

developedby: Also, if we want external people to be able to write these functions, we should probably encapsulate this with some nicer api for reading arguments from a tuple, just like we do for bytes and strings

developedby (changes requested):

src/run.c
enricozb: The `argm` passed to io functions is parsed through `readback_ctr`, so it is already expanded. I've added a `readback_tup`, I'll expose this in another PR.

src/run.c
```diff
@@ -345,11 +443,17 @@ Port io_get_time(Net* net, Book* book, Port argm) {
 // `argm` is a tuple (CON node) of the high and low
 // 24 bits for a 48-bit duration in nanoseconds.
 Port io_sleep(Net* net, Book* book, Port argm) {
+  Tup tup = readback_tup(net, book, argm, 2);
+  if (tup.elem_len != 2) {
+    fprintf(stderr, "io_sleep: expected 3-tuple\n");
```
developedby, line 448:
```suggestion
    fprintf(stderr, "io_sleep: expected 2-tuple\n");
```

developedby (changes requested):

developedby (approved):

PR #388, opened 2024-06-12
Correct an erratum in the paper

Somfic:
While reading the paper I noticed a possible erratum in *13.3 Single Core Inefficiency*.

> (…) will easily outperform an HVM2 one that uses thousands of threads, 
given the C version is doing no allocation, while ***C*** is allocating a 
tree-like recursive stack

I think it should be:

> (…) will easily outperform an HVM2 one that uses thousands of threads, 
given the C version is doing no allocation, while ***HVM2*** is allocating a 
tree-like recursive stack

Replacing *“C”* with *“HVM2”*.


enricozb:
Thanks! fixed by: https://github.com/HigherOrderCO/HVM/pull/412

PR #391, opened 2024-06-18
tan + sin funcs

### LINKED ISSUE #384: Add trigonometric function operators for F24
**Is your feature request related to a problem? Please describe.**
A request by a Bend user in this issue: https://github.com/HigherOrderCO/Bend/issues/570.

We have log, atan2 and pow for floats, but not trigonometric functions like sin, cos and tan.
I think they're common enough to warrant being a native operation. We should have at least one of them, the others could be easily derived (cos(x) = sqrt(1 - sin(x)²), etc)

**Describe the solution you'd like**
Add at least one of `sin`, `cos` and `tan` functions as a native floating point operation. We could use `SHL` and `SHR` tags to implement them for f24 types.

**Describe alternatives you've considered**
Approximating them using the already present operations. I think it can be done in around ~10-20 interactions each.

enricozb:
fixes https://github.com/HigherOrderCO/HVM/issues/384


src/hvm.c
```diff
@@ -570,6 +570,10 @@ static inline Numb operate(Numb a, Numb b) {
         case OP_AND: return new_f24(atan2f(av, bv));
         case OP_OR:  return new_f24(logf(bv) / logf(av));
         case OP_XOR: return new_f24(powf(av, bv));
+        case OP_SHL: return new_f24(sin(av + bv));
+        case FP_SHL: return new_f24(sin(av + bv));
```
developedby, line 574: We don't need to waste f24 FP_SHL and FP_SHR here

src/hvm.cu
```diff
@@ -676,6 +676,10 @@ __device__ __host__ inline Numb operate(Numb a, Numb b) {
         case OP_AND: return new_f24(atan2f(av, bv));
         case OP_OR:  return new_f24(logf(bv) / logf(av));
         case OP_XOR: return new_f24(powf(av, bv));
+        case OP_SHL: return new_f24(sin(av + bv));
+        case FP_SHL: return new_f24(sin(av + bv));
```
developedby, line 680: Same

src/hvm.rs
```diff
@@ -407,6 +407,10 @@ impl Numb {
           OP_AND => Numb::new_f24(av.atan2(bv)),
           OP_OR  => Numb::new_f24(bv.log(av)),
           OP_XOR => Numb::new_f24(av.powf(bv)),
+          OP_SHL => Numb::new_f24((av + bv).sin()),
+          OP_SHR => Numb::new_f24((av + bv).tan()),
+          FP_SHL => Numb::new_f24((av + bv).sin()),
```
developedby, line 412: and here

nicolas-abril (changes requested):

src/ast.rs
```diff
@@ -79,6 +79,8 @@ impl<'i> CoreParser<'i> {
       _ if self.try_consume("&")   => hvm::OP_AND,
       _ if self.try_consume("|")   => hvm::OP_OR,
       _ if self.try_consume("^")   => hvm::OP_XOR,
+      _ if self.try_consume("sin") => hvm::OP_SHL,
```
developedby, line 82: These symbols are trying to uniquely map to operators, so we shouldn't add sin and tan here. Notice how there's no pow, log and atan2.

nicolas-abril (changes requested):

src/ast.rs
enricozb: fair enough but in the context of `[` nothing else could appear there, thought it would be a bit more readable.

nicolas-abril (approved):

PR #398, opened 2024-06-26
Use bool type from stdbool.h

developedby:
Defining `bool` ourselves can cause all sorts of problems. For example, if the implementation of the libraries we're using import another another definition of bool, HVM will not compile.

Or if a user tries to write FFI for HVM and includes a definition of bool.

One option was to `#define bool _Bool`, which is what stdbool does, but including the standard library is the more standard way of doing it.

Also, `hvm.cuh` tried to redefine `bool` in c++ which would cause an error.

Another effect is that we can now compile hvm with C23


Sipher:
Tested  this on my machine and worked

Macbook Pro M3. I was having trouble compiling HVM, now it is working smoothly

enricozb (approved):

PR #399, opened 2024-06-26
Implement missing fp_shl and fp_shr for float types

imaqtkatt:


developedby:
sin and tan are commutative so they shouldn't implement the FP_ operations

imaqtkatt:
got it, so bend is compiling this operation wrong since the code below reaches unreachable for me:

```bend
foo x = (<< x 0.0)

main = (foo 3.0)
```

developedby:
@imaqtkatt bit shifting operations are not defined for f24 data, since they wouldn't meaningful (the mantissa bits and the exponent bits would get mixed).

`(<< x 0.0)` should be a type error and `sin` and `tan` are defined via native hvm definitions in Bend, preventing the auto-generation of `FP_SHL` nodes

PR #400, opened 2024-07-01
Updates actions

kings177:


developedby:
Why is this needed?

kings177:
> Why is this needed?

I am moving our self-hosted runner to another machine, and also, it wasn't working properly before apparently, since there were some errors that it didn't catch,

PR #407, opened 2024-08-01
return Result<T, E> from IO functions

enricozb:


src/hvm.c
```diff
@@ -1234,7 +1234,7 @@ Port expand(Net* net, Book* book, Port port) {
   Port old = vars_load(net, get_val(ROOT));
   Port got = peek(net, port);
   while (get_tag(got) == REF) {
-    boot_redex(net, new_pair(new_port(REF,get_val(got)), ROOT));
```
developedby, line 1237: Is this just to avoid one deref interaction every iteration of the expand?

src/run.c
```diff
@@ -225,6 +231,64 @@ Port inject_bytes(Net* net, Bytes *bytes) {
...
+  }
+
+  u32 v1 = tm[0]->vloc[0];
+
+  u32 n1 = tm[0]->nloc[0];
+  u32 n2 = tm[0]->nloc[1];
+  u32 n3 = tm[0]->nloc[2];
+
+  vars_create(net, v1, NONE);
+  Port var = new_port(VAR, v1);
```
developedby, line 248: What happens if the error value is not a var, does it just cause a link interaction?

src/run.c
```diff
@@ -334,78 +399,84 @@ Port io_open(Net* net, Book* book, Port argm) {
...
     }
   }
 
-  fprintf(stderr, "io_open: too many open files\n");
-
   free(name.buf);
   free(mode.buf);
 
-  return new_port(ERA, 0);
+  fprintf(stderr, "io_open: returning ERR\n");
```
developedby, line 413: Should remove this print, right?

src/run.c
```diff
@@ -461,18 +533,19 @@ Port io_get_time(Net* net, Book* book, Port argm) {
...
   node_create(net, loc, new_pair(new_port(NUM, new_u24(time_hi)), new_port(NUM, new_u24(time_lo))));
-  // Return the encoded time
+
   return new_port(CON, loc);
 }
 
 // Sleeps.
 // `argm` is a tuple (CON node) of the high and low
 // 24 bits for a 48-bit duration in nanoseconds.
+// Returns: Result<*, *>
```
developedby, line 543: Since the only error case here is a type error, I think its fine not to return a Result. This is fine, but I'd just unwrap the value in Bend since it should never fail in a correct program

src/run.c
```diff
@@ -538,28 +618,31 @@ Port io_dl_call(Net* net, Book* book, Port argm) {
...
 // Closes a loaded dylib, reclaiming the handle.
+//
+// Returns:  Result<*, String | *>
 Port io_dl_close(Net* net, Book* book, Port argm) {
   void* dl = readback_dylib(argm);
   if (dl == NULL) {
     fprintf(stderr, "io_dl_close: invalid handle\n");
-    return new_port(ERA, 0);
+
+    return inject_err(net, new_port(ERA, 0));
```
developedby, line 635: Can you return a string instead of an Era in this case?
I feel like this should be a recoverable/debuggable error

nicolas-abril (changes requested):
Remember to also update hvm.h

src/run.c
enricozb: We discussed returning a result here, since `call` will return a `Result<T, IOError<E>>`, as the name or args may be invalid when passed to `call`.

src/hvm.c
enricozb: Since `get_tag(got) == REF`, we have `new_port(REF,get_val(got)) == got`.

src/run.c
enricozb: The var `v1` is used to link two aux ports. The returned net is:
```
((RESULT_ERR (err x)) x)
```
where `RESULT_ERR` is the numeric tag of the `Result/Err` variant, and the `x` var in the snippet above is `v1` in the code.

src/run.c
enricozb: I'm going to return an `IOError/Type` meaning an invalid argument.

examples/demo_io/main.bend
```diff
@@ -1,63 +1,35 @@
...
+  match res:
+    case Result/Ok:
+      return res.val
+    case Result/Err:
+      return res.val
 
-def IO_T/MAGIC:
-  return (0xD0CA11, 0xFF1FF1)
+def open():
+  return call("OPEN", ("./README.md", "r"))
```
developedby, line 11: Opening the readme is good as an example, but when running as a test it will error if we change the start of the readme.

Maybe the program reading its own source code would be better

nicolas-abril (changes requested):
Remember to update the lib headers `hvm.h` and `hvm.cuh`

nicolas-abril (approved):
looks alright

enricozb:
Going to hold off on merging as there is some issue with the C runtime hanging on this code
```
def unwrap(res):
  match res:
    case Result/Ok:
      return res.val
    case Result/Err:
      return res.val

(Main) =
  with IO {
    ask res = (call "foo" (1, 2))

    (unwrap res)
  }
```

Issue #309, opened 2024-05-19
hvm-c segfaults with some programs


developedby:
**Describe the bug**
Some programs are causing hvm-c to run into a segmentation fault.
As reported in https://github.com/HigherOrderCO/Bend/issues/397.

**To Reproduce**
Run this program with `bend run-c` https://github.com/HigherOrderCO/Bend/files/15368994/sat.bend.txt
Or the compiled-to-hvm version with `hvm run-c` https://github.com/HigherOrderCO/Bend/files/15368996/sat.hvm.txt

Another user reported on discord a segfault when running a simple hello world with `bend run-c`, but I haven't been able to reproduce
```
def main():
  return "Hello, world!"
```
User says: "I am running on WSL2 under Windows 10"

Janiczek:
Another case of SIGSEGV in `run-c`, on Mac M1 Pro:
```py
def testArrayMapArrayNonrecursive(_):
  * = lambda x: x+1
  return 1

def main():
  return 1
```

Interestingly, either removing the lambda definition or renaming the test makes the error go away.

The renaming behaviour makes me think this might be related to https://github.com/HigherOrderCO/HVM/pull/305

zylkowski:
I get SIGSEGV no matter the bend script I run. Any program I wrote and ran with `bend run-c` resulted in SIGSEGV.

OS: WSL2 under Windows 10
CPU: Ryzen 9 5950x

VictorTaelin:
This one was fixed with larger names I believe, right?

enricozb:
Closing as I can't reprocude. If a segfault comes up again another issue should be opened.

sinsikyan:
I have this issue.

OS: ubuntu on WSL2 on win11
GPU: nividia geforece 940MX

hvm 2.0.22
bend-lang 0.2.36

nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Fri_Jan__6_16:45:21_PST_2023
Cuda compilation tools, release 12.0, V12.0.140
Build cuda_12.0.r12.0/compiler.32267302_0

gcc (Ubuntu 12.3.0-17ubuntu1) 12.3.0

g++ (Ubuntu 12.3.0-17ubuntu1) 12.3.0

`hvm run` runs perfectly
`hvm run-c` Segmentation fault on every program including
```
def main:
  return "test"
```
i compiled it to .hvm first then run it using hvm of cause.

[...]

Issue #314, opened 2024-05-20
Bitonic Sort on fails on CUDA with (error code an illegal memory access was encountered)


nicolas-abril:
Originally from https://github.com/HigherOrderCO/Bend/issues/364 by user @ethanbarry

Description

When I run the compiled CUDA [bitonic sorter example](https://gist.github.com/VictorTaelin/face210ca4bc30d96b2d5980278d3921) (linked in the README) I get this error:

Failed to launch kernels (error code an illegal memory access was encountered)!

To Reproduce

    bend gen-cu sorter.bend > sorter.cu
    nvcc sorter.cu -o sorter
    prime-run ./sorter (Launches it on the GPU for Arch Linux.)
    Error recieved.

Expected behavior

The program runs on the GPU.
Desktop (please complete the following information):

    OS: Linux (Arch 6.9.1-arch1-1)
    CPU: Intel i7-11800H
    GPU: RTX 3050 Ti Mobile
    GPU Driver: Nvidia open kernel modules v550.78
    CUDA release 12.4, V12.4.131

Additional context

The program runs using the C codegen backend, but with the CUDA backend, it seems to fail regardless of what I do. If anyone is curious about the prime-run command, it's really just a script that forces the dGPU to handle a task - nothing fancy.

NotCyberLemon:
I am here from the main Bend repo with issue  #[Bitonic Sort example failed with GPU kernel error.](https://github.com/HigherOrderCO/Bend/issues/364)

I too am having a kernel memory issue:

```sh
$ ./sorter # The same as prime-run due to environment variables already being set.
| Failed to launch kernels (error code an illegal memory access was encountered)!
```

I am also running a mobile gpu where I am getting this issue.

Some GPU properties and info [*from exec*](https://gist.github.com/NotCyberLemon/5f0daef77b44334c7eae210c50fa4d71):

```
--- General Information for device 0 ---
Name: NVIDIA GeForce RTX 3060 Laptop GPU
Compute capability: 8.6
Clock rate: 1425000
Device copy overlap: Enabled
Kernel execution timeout: Enabled

--- Memory Information for device 0 ---
Total global memory: 5996544000
Total constant memory: 65536
Max memory pitch: 2147483647
Texture alignment: 512

--- MP Information for device 0 ---
Multiprocessor count: 30
Shared memory per MP: 49152
Registers per MP: 65536
Threads in warp: 32
Max threads per block: 1024
[...]

PR #413, opened 2024-08-16
adds dynamic shared mem allocation to cuda kernels

### LINKED ISSUE #283: CUDA not available or Failed to Launch Kernels (error code invalid argument)
References: https://github.com/HigherOrderCO/Bend/issues/320 by rubenjr0 with contents:

"
Hello! I think I've encountered a bug. When running this example from the readme:

```py
def sum(depth, x):
  switch depth:
    case 0:
      return x
    case _:
      fst = sum(depth-1, x*2+0) # adds the fst half
      snd = sum(depth-1, x*2+1) # adds the snd half
      return fst + snd
    
def main:
  return sum(30, 0)
```

The output is 0. I've tried `bend run`, `bend run-c`, and `bend gen-cu` (bend run-cu says cuda is not available, so I manually compile it with nvcc).

The output on my machine when running `sum(24, 0)` is 8388608, but on equivalent Haskell and Python programs the programs return 140737479966720. The results start to diverge when depth>=13.

I was wondering what could be causing these issues, both the incorrect result when depth>=13, and the result=0 when depth>=25.

My computer specs:

OS: Pop_OS 22.04
CPU: AMD Ryzen 5 2600x (12 cores)
GPU: NVIDIA RTX 4060 ti (16GB)
"

kings177:
with this hvm.cu should work virtually on every NVIDIA GPU out there (assuming 5.0 CC and above). it dynamically allocates shared memory based on the GPU's capabilities, specifically 3072 less bytes than the max opt-in shared mem available, as some shared arrays use roughly (a little bit less than) that amount of shared mem.

since shared mem allocation needs to be known at compile time, `get_shared_mem.cu` calculates the available shared mem in build time, which is ran by `build.rs` that then generates a header file `shared_mem_config.h` with the correct hex value for the local net.

Closes: #283 and #314 (supposedly)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,7 +9,6 @@ fn main() {
   println!("cargo:rerun-if-changed=src/run.cu");
   println!("cargo:rerun-if-changed=src/hvm.cu");
   println!("cargo:rerun-if-changed=src/get_shared_mem.cu");
-  println!("cargo:rerun-if-changed=src/shared_mem_config.h");
   println!("cargo:rustc-link-arg=-rdynamic");
 
   match cc::Build::new()
@@ -35,28 +34,31 @@ fn main() {
     }
 
     // Compile get_shared_mem.cu
-    if let Ok(output) = Command::new("nvcc")
+    let shared_mem_value = Command::new("nvcc")
       .args(&["src/get_shared_mem.cu", "-o", "get_shared_mem"])
       .output()
-      .and_then(|_| Command::new("./get_shared_mem").output()) {
-        if output.status.success() {
-          let shared_mem_str = String::from_utf8_lossy(&output.stdout).trim().to_string();
-          std::fs::write("src/shared_mem_config.h", format!("#define HVM_SHARED_MEM {}", shared_mem_str))
-            .expect("Failed to write shared_mem_config.h");
-        } else {
-          println!("cargo:warning=\x1b[1m\x1b[31mWARNING: Failed to get shared memory size. Using default value.\x1b[0m");
-        }
-    } else {
-      println!("cargo:warning=\x1b[1m\x1b[31mWARNING: Failed to compile or run get_shared_mem.cu. Using default shared memory value.\x1b[0m");
-    }
+      .and_then(|_| Command::new("./get_shared_mem").output())
+      .ok()
+      .and_then(|output| if output.status.success() {
+          Some(String::from_utf8_lossy(&output.stdout).trim().to_string())
+      } else {
+        None
+      })
+      .unwrap_or_else(|| {
+        println!("cargo:warning=\x1b[1m\x1b[31mWARNING: Failed to get shared memory size. Using default value.\x1b[0m");
+        "0x2000".to_string()
+      });
 
     // Clean up temporary executable
     let _ = std::fs::remove_file("get_shared_mem");
+
+    println!("cargo:warning=\x1b[1m\x1b[33mShared memory size set to: {}\x1b[0m", shared_mem_value);
 
     cc::Build::new()
       .cuda(true)
       .file("src/run.cu")
       .define("IO", None)
+      .define("HVM_SHARED_MEM", Some(shared_mem_value.as_str()))
       .flag("-diag-suppress=177") // variable was declared but never referenced
       .flag("-diag-suppress=550") // variable was set but never used
       .flag("-diag-suppress=20039") // a __host__ function redeclared with __device__, hence treated as a __host__ __device__ function
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,5 @@
+use std::process::Command;
+
 fn main() {
   let cores = num_cpus::get();
   let tpcl2 = (cores as f64).log2().floor() as u32;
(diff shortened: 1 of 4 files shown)


src/get_shared_mem.cu
```diff
@@ -0,0 +1,24 @@
...
+    cudaGetDeviceProperties(&prop, device);
+
+    size_t sharedMemPerBlock = prop.sharedMemPerBlock;
+    int maxSharedMemPerBlockOptin;
+    cudaDeviceGetAttribute(&maxSharedMemPerBlockOptin, cudaDevAttrMaxSharedMemoryPerBlockOptin, device);
+
+    size_t maxSharedMem = (sharedMemPerBlock > (size_t)maxSharedMemPerBlockOptin) ? sharedMemPerBlock : (size_t)maxSharedMemPerBlockOptin;
+
+    // Subtract 3KB (3072 bytes) from the max shared memory as is allocated somewhere else
+    maxSharedMem -= 3072;
```
enricozb, line 16: How did you compute this? I think this is from the automatic variables in some of the kernels. I think overall this is a fine approach for now but if we change the 3KB alloc in the runtime this will have to change too.

enricozb (approved):

src/hvm.cu
```diff
@@ -126,9 +126,15 @@ struct RBag {
...
+#include "shared_mem_config.h"
+
+#ifndef HVM_SHARED_MEM
+#define HVM_SHARED_MEM 0x2000 // Default value
+#endif
+
 // Local Net
-const u32 L_NODE_LEN = 0x2000;
-const u32 L_VARS_LEN = 0x2000;
+const u32 L_NODE_LEN = HVM_SHARED_MEM;
```
developedby, line 136: Aren't there places in the code that expect L_NODE_LEN to have this exact value?

If you change it I think we'll start to get either memory leaks, out of bounds access or non local memory use.

build.rs
```diff
@@ -23,13 +27,32 @@ fn main() {
...
     }
 
+    // Compile get_shared_mem.cu
+    if let Ok(output) = Command::new("nvcc")
+      .args(&["src/get_shared_mem.cu", "-o", "get_shared_mem"])
+      .output()
+      .and_then(|_| Command::new("./get_shared_mem").output()) {
+        if output.status.success() {
+          let shared_mem_str = String::from_utf8_lossy(&output.stdout).trim().to_string();
+          std::fs::write("src/shared_mem_config.h", format!("#define HVM_SHARED_MEM {}", shared_mem_str))
```
developedby, line 44: Instead of writing it to a build generated file, can't we just set a define flag, like we do for the number of cores? It's prpbably possible with nvcc as well

src/hvm.cu
developedby: Also, if this number is smaller then default, the programs compiled by bend won't run.

If it's smaller the performance will also tank incredibly fast. (see the issue where someone halved this number and got worse performamce than in the cpu)

nicolas-abril (changes requested):

developedby:
Probably closes #309

src/get_shared_mem.cu
kings177: the sum of the alloc size of some local shared arrays is roughly ~3KB, i know that this is not the perfect way to do this, but it works for now.

src/hvm.cu
kings177: > Aren't there places in the code that expect L_NODE_LEN to have this exact value?

i wouldn't say that it expects to have this value, but rather that it expects to have the numbers of a device with 8.9 of compute capability, but then again, this is not a general optimization of the CUDA version, this is just to make the allocation of shared mem on local net dynamic, so that users can install the hvm without having to manually change it on the `hvm.cu`, the rest of the code is still the same, made with the numbers of a 4090 in mind. We can slowly improve the code to make it `device` based, instead of hard-coded to the specs of a `4090`.

> Also, if this number is smaller then default, the programs compiled by bend won't run.

can you show me an example of what was ran and what was the device/numbers used when this happened?


> If it's smaller the performance will also tank incredibly fast. (see the issue where someone halved this number and got worse performamce than in the cpu)

i mean, that's of course, if you give it less memory, it will have less memory, besides the fact that the rest of the code is 'optimized' for a `4090`.

we can like, whenever someone installs it using a GPU with <99KB of max shared mem per block, we give a warning saying something along the lines of: 
"HVM is currently optimized to run on devices with >=96KB of max shared mem, please be aware that your GPU performance will be reduced dramatically"

build.rs
kings177: true, that sounds cleaner indeed, will change it to that

src/hvm.cu
developedby: > can you show me an example of what was ran and what was the device/numbers used when this happened?



> > Also, if this number is smaller then default, the programs compiled by bend won't run.
> 
> can you show me an example of what was ran and what was the device/numbers used when this happened?

The number of nodes in the local buffer determines the maximum size of hvm definitions. If we decrease `L_NODE_LEN` then Bend also needs to generate programs with smaller maximum definition size.

Issue #420, opened 2024-08-20
Add IO function for creating new directories


kings177:
Implement a new IO function in the HVM to give Bend the ability to create new directories in the file system.

## 1. `create_directory(path, parents=False, exist_ok=False)`
```rs
@spec create_directory(str, bool, bool) -> Result[None, Error]
```

Creates a new directory at the specified path.

### Parameters
- `path`: The path where the new directory should be created
- `parents`: If `True`, create parent directories as needed. If `False`, parent directories must exist
- `exist_ok`: If `True`, don't raise an error if the directory already exists

Returns `Ok(None)` if successful, or `Err(reason)` if an error occurs.

Possible (but not limited to) errors:
- `FileExists`: The directory already exists and `exist_ok` is `False`
- `FileNotFound`: A parent directory doesn't exist and parents is `False`
- `PermissionDenied`: The user doesn't have permission to create the directory
- `OSError`: Other OS-level errors (e.g., disk full, invalid path)

### Examples
```py
# Create a new directory
result = create_directory("new_folder")
# Ok(None)
```

```py
# Create a directory with non-existent parents, creating them as needed
result = create_directory("path/to/new/folder", parents=True)
# Ok(None)
```

```py
# Create a directory, allowing it to exist
result = create_directory("maybe_existing_folder",  exist_ok=True)
# Ok(None)
```

```py
# Try to create an existing directory
result = create_directory("existing_folder")
# Err(FileExists)
```

```py
# Create a directory with non-existent parents
result = create_directory("path/to/new/folder")
# Err(FileNotFound)
```

## Considerations
- Implement proper error handling and propagation.
- Ensure thread-safety for directory creation.
- Handle path normalization (e.g., handling '...' and '.' in paths).
- Consider preserving or setting specific permissions for newly created directories.
- Be aware of and handle potential race conditions (e.g., directories being created/deleted concurrently)
- Implement appropriate logging for debugging.

## Test cases to implement
1. Create a new directory in an existing parent directory
2. Attempt to create a directory that already exists (with and without `exist_ok`)
3. Create a directory with multiple levels of non-existent parent directories (with and without `parents` flag)
4. Attempt to create a directory without proper permissions
5. Create a directory with a very long path name (test path length limits)
6. Create a directory with special characters in the name
7. Attempt to create a directory in a read-only file system
8. Create multiple directories concurrently to test thread-safety
9. Create a directory where a file with the same name already exists
[...]

Issue #418, opened 2024-08-20
Add IO functions for removing files and directories


kings177:
Implement a file deletion functionality to give Bend the ability to delete files from the filesystem. This should include both single file deletion and recursive directory deletion.

## 1. `delete_file(path)`
```rs
@spec delete_file(str) -> Result[None, Error]
```
Deletes a single file or an empty directory at the specified path.

Returns `Ok(None)` if successful, or `Err(reason)` if an error occurs.

This function attempts to remove both files and empty directories without first checking the type of the path. It tries to remove the path as a file first, and if that fails, it attempts to remove it as a directory.

Possible (but not limited to) errors:
 - `FileNotFound`: The file or directory does not exist
 - `PermissionDenied`: Lack of permission to delete the file or directory
 - `NotEmpty`: Attempted to remove a non-empty directory
 - `OSError`: Other OS-level errors (e.g., I/O error)

### Examples
```py
# Delete an existing file
result = delete_file("example.txt")
# Ok(None)
```
```py
# Delete an empty directory
result = delete_file("empty_dir")
# Ok(None)
```
```py
# Try to delete a non-existent file or directory
result = delete_file("nonexistent.txt")
# Err(FileNotFound)
```
```py
# Try to delete a non-empty directory
result = delete_file("my_directory")
# Err(NotEmpty)
```

## 2. `delete_directory(path, recursive=False)`
```rs
@spec delete_directory(str, bool) -> Result[None, Error]
```
Deletes a directory at the specified path. If recursive is `True`, it will delete the directory and all its contents.

Returns Ok(None) if successful, or Err(reason) if an error occurs.

Note: For non-recursive deletion of an empty directory, this function behaves the same as `delete_file(path)`.

Possible (but not limited to) errors:
- `FileNotFound`: The directory does not exist
- `PermissionDenied`: Lack of permission to delete the directory or its contents
- `NotADirectory`: The path is not a directory
- `OSError`: Other OS-level errors (e.g., I/O error)

### Examples
```py
# Delete an empty directory
result = delete_directory("empty_dir")
# Ok(None)
```
```py
# Try to delete a non-empty directory
result = delete_directory("non_empty_dir")
# Err(NotEmpty)
```
```py
# Delete a directory and its contents
result = delete_directory("non_empty_dir", recursive=True)
# Ok(None)
```
```py
# Try to delete a file using delete_directory
result = delete_directory("file.txt")
# Err(NotADirectory)
```

## Considerations
[...]

PR #426, opened 2024-08-28
Implement remove, remove_all and mkdir

imaqtkatt:
Despite the branch name it implements 3 new functions related to the removal of files or directories and the creation of directories


nicolas-abril:
Implements both #418 and #420

src/run.cu
```diff
@@ -834,6 +838,121 @@ Port io_dl_close(GNet* gnet, Book* book, Port argm) {
...
+// Returns Ok(None) if sucessfull, or Err(reason) if an error occurs.
+// Returns: Result<*, IOError<i24>>
+Port io_mkdir(GNet* gnet, Port argm) {
+  Str name = gnet_readback_str(gnet, argm);
+
+  const mode_t mode = 0777;
+  int result = mkdir(name.buf, mode);
+  free(name.buf);
+
+  if (result) {
```
enricozb: Can we keep the var names the same? In the C runtime you called this `status`.

enricozb (approved):

src/run.cu
```diff
@@ -846,6 +965,9 @@ void book_init(Book* book) {
   book->ffns_buf[book->ffns_len++] = (FFn){"DL_OPEN", io_dl_open};
   book->ffns_buf[book->ffns_len++] = (FFn){"DL_CALL", io_dl_call};
   book->ffns_buf[book->ffns_len++] = (FFn){"DL_CLOSE", io_dl_open};
+  book->ffns_buf[book->ffns_len++] = (FFn){"DELETE_FILE", io_delete_file};
+  book->ffns_buf[book->ffns_len++] = (FFn){"DELETE_DIRECTORY", io_delete_directory};
```
nicolas-abril: Following the convention of using the C/unix name for these functions, it would be better to call them "RM" and "RMDIR".

Also, if we're using "rmdir" we definitely should not also use "delete_directory", either go with one style or the other.

HVM in general prefers short names

tests/programs/io/create_directory.bend
```diff
@@ -0,0 +1,21 @@
+#{
+  Creates the batata directory and then deletes it.
+#}
+
+test-io = 1
+
+IO/FS/mkdir path =
```
nicolas-abril: This test will break once Bend adds functions with these names in the prelude (which we have been doing for now)

src/run.c
```diff
@@ -716,6 +719,120 @@ Port io_dl_close(Net* net, Book* book, Port argm) {
...
+      len = path_len + strlen(p->d_name) + 2;
+      buf = malloc(len);
+
+      if (buf) {
+        struct stat statbuf;
+        snprintf(buf, len, "%s/%s", path, p->d_name);
+
+        if (!stat(buf, &statbuf)) {
+          if (S_ISDIR(statbuf.st_mode)) {
+            r2 = delete_directory_recursive(buf);
```
nicolas-abril: This will stack overflow. Prefer using `nftw` or a stack/list of files to remove

nicolas-abril (changes requested):

nicolas-abril:
I've also added some comments on the specs for this issue, I think we need to change the interfaces a little

src/run.cu
```diff
@@ -1,6 +1,10 @@
 #include <dlfcn.h>
 #include <errno.h>
 #include <stdio.h>
+#include <dirent.h>
+#include <sys/stat.h>
+#include <string.h>
```
nicolas-abril: Is `string.h` still being used? I don't see it in the changes

build.rs
```diff
@@ -37,6 +37,8 @@ fn main() {
       .flag("-diag-suppress=177") // variable was declared but never referenced
       .flag("-diag-suppress=550") // variable was set but never used
       .flag("-diag-suppress=20039") // a __host__ function redeclared with __device__, hence treated as a __host__ __device__ function
+      .flag("-diag-suppress=68") // integer conversion resulted in a change of sign
```
nicolas-abril: This does not look right, what is triggering this warning, specifically?

build.rs
```diff
@@ -37,6 +37,8 @@ fn main() {
...
+      .flag("-diag-suppress=2464") // conversion from a string literal to "char *" is deprecated
```
nicolas-abril: I think you can remove this warning now that you don't mess with paths manually, right?

nicolas-abril (changes requested):

build.rs
imaqtkatt: `ctr.tag = -1` triggers this warning

nicolas-abril: Was this a warning we already had? I don't see that in your changes

src/run.cu
imaqtkatt: I know `strlen` is being called somewhere but jump to definition is not working for me in this file

build.rs
imaqtkatt: I think so

src/run.cu
nicolas-abril: You're right, i don't understand how it works without it. I guess some gnu extensions?

PR #433, opened 2024-10-24
paper Switch 2 typo fix

DKormann:
I believe there has been a slight typo.
not sure how to compile the pdf though


DKormann:
i think the correct rule should be something like this right
```
#n+1 ~ ?(A B)`
_____________
A ~ (* x)
#n ~ ?(x B)
```

VictorTaelin:

PR #434, opened 2024-11-05
fix double negative

pepijndevos:


VictorTaelin:

PR #435, opened 2025-01-29
Interaction Net graph dumping and visualization

lukasstraub2:
Hello, we are a student project from the University of Stuttgart and are evaluating Bend.

We started out with implementing some classical algorithms as well as a matrix multiplication to get familiar with Bend. Since Bend is relatively new, instead of asessing absolute performance, we focused on scalability and parallelism instead.

But soon we hit a roadblock: None of our programs "felt" parallel, even programs that should parallelize well in Bend like matrix multiplication or megesort. And worse, there where no indications or debugging tools to tell if the programs really where parallel and, if not, why.

It was clear that we needed proper tools. So we decided to modify the HVM2 rust interpreter to dump the Interaction Net graph in graphviz dot format at every step. We also quickly hacked together a viewer in html and javascript to be able to step trough the graph dumps.

Our modification is able to dump the graph in sequential mode (only one redex is evaluated in each step) or in parallel mode (all redexes are evaluated in each step). In parallel mode, it also calculates the average number of regexes in each step, thus giving a measurement of how parallel a given bend program is.

You can see some Interaction Nets here, use the wasd Keys to step through the visualizations: http://w3studi.informatik.uni-stuttgart.de/~straubls/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ examples/**/main
 examples/**/*.c
 examples/**/*.cu
 .out.hvm
+dots.js
 
 # nix-direnv
 /.direnv/
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ examples/**/*.c
 examples/**/*.cu
 .out.hvm
 dots.js
+profile.csv
 
 # nix-direnv
 /.direnv/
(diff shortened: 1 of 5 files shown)


wyattgill9:
Every website visualization is the same.

VictorTaelin:

